PR #19, opened 2024-01-02
oci scaffolding in agent

autodidaddict:
this adds _just_ the agent functionality. There are some corresponding changes that need to be made on the node.

I think we also want to make sure that you can configure which workloads a node is allowed to run. I don't think we want people to use the "fat" rootfs unless they're explicitly planning on running docker images.

We should also probably document that running docker images like this isn't the most efficient use of resources.


autodidaddict:
This is now too out of date to make sense trying to merge. The current plan is to evaluate a lightweight OCI solution that doesn't require us manually using the docker daemon

PR #28, opened 2024-01-04
Adds forensic (aka messy toddler) mode and nex extract

autodidaddict:
1. Adds the "forensic" flag to `nex-node`, which will prevent it from deleting the rootfs of a firecracker VM that has shut down
2. Adds the `extract` option to the nex-cli, which will attempt to pull all relevant logs out of a rootfs `ext4` image.


autodidaddict:
Going to close this soon as we've since found a better, more developer-friendly way to do this

PR #78, opened 2024-01-22
Upgrade/build params

jordan-rash:
This allows for the nex cli to build for more OSs ensuring that node subcommand and ui only build when approperate


jordan-rash:
the `goreleaser` added to the CI in this PR only does the build.  Its just there to give a heads up it might break the release

PR #88, opened 2024-01-25
initial support for using nats contexts in the nex cli

autodidaddict:
Note that I've only tried this on macOS. I'll try and switch over to Linux and verify tomorrow


internal/node/init.go
```diff
@@ -54,10 +55,11 @@ func CmdUp(opts *nexmodels.Options, nodeopts *nexmodels.NodeOptions, ctx context
 	}
 }
 
+// TODO
+// FIXME
+// if I export this to the `nex` binary, then macOS won't build.
+// why? what did I do to deserve this?
```
jordan-rash, line 61: 🤣

jordan-rash (approved):
Changes look good to me.

PR #92, opened 2024-01-25
RootFS Docs Update

jordan-rash:
This makes it more clear on how to build the rootfs locally

### DIFF
--- a/agent/fc-image/README.md
+++ b/agent/fc-image/README.md
@@ -8,7 +8,7 @@ The files in this directory are what are used to generate the default `rootfs.ex
 
 ### To Generate File
 
-> NOTE: You will be running `sudo go` which will be using root's PATH to find the go binary.  This isn't how typical go installations are done, so you will need to check `sudo which go` and see if the binary is found.  If not, we have found that soft linking it is the easiest solution.  `sudo ln -s /path/to/go/binary/go /usr/local/bin/go`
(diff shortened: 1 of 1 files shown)


agent/fc-image/README.md
```diff
@@ -8,6 +8,8 @@ The files in this directory are what are used to generate the default `rootfs.ex
 
 ### To Generate File
 
-`sudo go run .`
+> NOTE: You will be running `sudo go` which will be using root's PATH to find the go binary.  This isn't how typical go installations are done, so you will need to check `sudo which go` and see if the binary is found.  If not, we have found that soft linking it is the easiest solution.  `sudo ln -s /path/to/go/binary/go /usr/local/bin/go`
```
autodidaddict, line 11: Might want to also mention here that this could still fail if root can't find the go path. I couldn't get it to work until I created a `/root/go` directory .. there's probably an environment variable that'll let us set that, right?

jordan-rash: Yea, i think we need to maybe bring this up a level TBH.  The tests also require go running as root, so this will be a problem there too. @kthomas what do you think?

kthomas: Locally I did indeed create `/root/go` but interestingly that didn't seem necessary in CI.

kthomas (dismissed):

kthomas (approved):

PR #97, opened 2024-01-26
Stop calling panic from node up and preflight

kthomas:
This PR also includes some initial refactoring to make the `MachineManager` more... manageable.


spec/node_test.go
```diff
@@ -59,9 +59,9 @@ var _ = Describe("nex node", func() {
 			})
 
 			It("should panic", func(ctx SpecContext) {
```
jordan-rash, line 61: should this read "should panic"??? probably "should error"

jordan-rash (changes requested):
Other then the test descriptions, looks good to me

jordan-rash (approved):
LGTM

PR #105, opened 2024-02-01
Install script easy button

jordan-rash:
Install script to keep things breezy


kthomas:
I think we should try to minimize the number of root directories. We could perhaps create a single `ops/` directory and put things like `_otel/` and `install.sh` underneath to keep things clean.

_install/install.sh
```diff
@@ -0,0 +1,90 @@
...
+downloadFolder="${TMPDIR:-/tmp}"
+mkdir -p ${downloadFolder}
+downloaded_file="${downloadFolder}/nex"
+executable_folder="/usr/local/bin"
+
+echo "[1/3] Download ${asset_uri} to ${downloadFolder}"
+rm -f ${downloaded_file}
+curl --silent --fail --location --output "${downloaded_file}" "${asset_uri}"
+
+echo "[2/3] Install nex to the ${executable_folder}"
```
autodidaddict, line 74: nit: probably don't need "the" here

autodidaddict (dismissed):

kthomas (approved):

PR #109, opened 2024-02-02
Fix ginkgo stability in github actions

kthomas:
After upgrading ginkgo version earlier this afternoon it appears github actions runners have become unstable, passing on the `feat/node-rebased` branch but then failing after being merged to `main`.


kthomas:
Moved this to draft status until we get to the bottom of it.

PR #110, opened 2024-02-03
Publish node stopped message prior to draining connection

kthomas:


kthomas:
See https://github.com/synadia-io/nex/pull/112

PR #111, opened 2024-02-03
Use noop meter provider when otel metrics are disabled

kthomas:


kthomas:
See https://github.com/synadia-io/nex/pull/112

PR #114, opened 2024-02-05
Downgrade OTel

jordan-rash:
Need this to get us at peace with docker version of OTel per this issue https://github.com/open-telemetry/opentelemetry-go/issues/2341


internal/node/telemetry.go
```diff
@@ -140,14 +141,15 @@ func (t *Telemetry) initMeterProvider() error {
 				semconv.SchemaURL,
 				semconv.ServiceName(t.serviceName),
 				semconv.ServiceVersion(VERSION),
+				attribute.String("node_id", "1234"), // FIXME-- this should be a unique identifier for the node
```
jordan-rash, line 144: @kthomas I need access to the node public key...how would be the best way to get that here?

internal/node/node.go
```diff
@@ -335,6 +335,7 @@ func (n *Node) shutdown() {
 		n.ncint.Close()
 		n.natsint.Shutdown()
 		n.natsint.WaitForShutdown()
+		_ = n.telemetry.providerShutdown(n.ctx)
```
kthomas, line 338: I would create a `Shutdown()` function on `Telemetry` itself and call `t.providerShutdown()` from there.

internal/node/telemetry.go
```diff
@@ -44,14 +47,16 @@ type Telemetry struct {
 	functionRunTimeNano    metric.Int64Counter
 }
 
-func NewTelemetry(log *slog.Logger, config *NodeConfiguration) (*Telemetry, error) {
+func NewTelemetry(log *slog.Logger, config *NodeConfiguration, nodePubKey string) (*Telemetry, error) {
```
kthomas, line 50: Probably should pass the `ctx` in as the first arg to `NewTelemetry` as well.

internal/node/telemetry.go
```diff
@@ -160,11 +167,10 @@ func (t *Telemetry) initMeterProvider() error {
...
-			if err := meterProvider.(*metricsdk.MeterProvider).Shutdown(context.Background()); err != nil {
-				t.log.Error("failed to shutdown OTel meter provider", slog.Any("err", err))
-			}
-		}()
+		if _, ok := meterProvider.(*metricsdk.MeterProvider); ok {
+			t.providerShutdown = meterProvider.(*metricsdk.MeterProvider).Shutdown
+		}
+
 	} else {
 		meterProvider = noop.NewMeterProvider()
```
kthomas, line 175: Should we keep a reference around to `meterProvider` instead of `providerShutdown`? Then in `Shutdown()` we can just call `t.meterProvider.Shutdown(ctx)`?

kthomas (approved):

Issue #40, opened 2024-01-09
node up should always start the first vm on the first available IP in the CIDR


autodidaddict:
Right now node up just continues allocating IP addresses. In production it won't take long for the IPs to run out.

Note that this behavior will happen regardless of how many times the node is started. Basically after 254 VMs from a fresh start, we're going to run out. So I see the requirements as follows:

* Start the node fresh with an empty /var/lib/cni so the first VM gets the `.2` address
* Before putting the 253rd VM into the pool, we should reset /var/lib/cni so that the next VM gets `.2`

kthomas:
Closed by https://github.com/synadia-io/nex/pull/115

PR #115, opened 2024-02-05
Additional testing and incremental refactor

kthomas:
This PR provides testing of the deployment of each of the workload types.

Some additional incremental refactoring can also be expected.

Closes https://github.com/synadia-io/nex/issues/40.

### DIFF
--- a/internal/node/config.go
+++ b/internal/node/config.go
@@ -161,4 +161,3 @@ func LoadNodeConfiguration(configFilepath string) (*NodeConfiguration, error) {
 
 	return &config, nil
 }
-
--- a/internal/node/config.go
+++ b/internal/node/config.go
@@ -24,21 +24,20 @@ var (
 // Node configuration is used to configure the node process as well
 // as the virtual machines it produces
 type NodeConfiguration struct {
-	DefaultResourceDir string `json:"default_resource_dir"`
-	KernelFilepath     string `json:"kernel_filepath"`
-	RootFsFilepath     string `json:"rootfs_filepath"`
-
 	CNI                 CNIDefinition     `json:"cni"`
-	ForensicMode        bool              `json:"-"`
+	DefaultResourceDir  string            `json:"default_resource_dir"`
 	ForceDepInstall     bool              `json:"-"`
 	InternalNodeHost    *string           `json:"internal_node_host,omitempty"`
 	InternalNodePort    *int              `json:"internal_node_port"`
+	KernelFilepath      string            `json:"kernel_filepath"`
 	MachinePoolSize     int               `json:"machine_pool_size"`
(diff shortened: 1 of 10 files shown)


spec/spec_suite_test.go
```diff
@@ -99,9 +99,34 @@ func startNATS(storeDir string) (*server.Server, *nats.Conn, *int, error) {
...
+		Storage:     nats.MemoryStorage,
+	})
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to create jetstream object store: %s", err)
+	}
+
 	return ns, nc, &port, nil
 }
 
+// return a new NATS connection to the server running on `_fixtures.natsPort`
```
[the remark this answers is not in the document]
kthomas: Removed 👍🏻

internal/node/config.go
```diff
@@ -49,6 +48,10 @@ type NodeConfiguration struct {
 func (c *NodeConfiguration) Validate() bool {
 	c.Errors = make([]error, 0)
 
+	if c.MachinePoolSize < 1 {
```
jordan-rash, line 51: instead of erroring here, can we warn and set it to 1?

kthomas: Not ideal IMHO, means you can have a malformed config and get away with it.

kthomas: Error gives the user a chance to fix their faulty config..

internal/node/machine_mgr.go
```diff
@@ -112,31 +113,40 @@ func (m *MachineManager) Start() {
...
+				continue
+			}
+
 			vm, err := createAndStartVM(m.ctx, m.config, m.log)
 			if err != nil {
 				m.log.Warn("Failed to create VMM for warming pool.", slog.Any("err", err))
 				continue
 			}
 
 			go m.awaitHandshake(vm.vmmID)
```
jordan-rash, line 143: if this go routine is non-blocking, the next line adds the VM to the map without verifying a handshake. Is that the behavior we want?

kthomas: Yes because we need to be able to find it later to shut it down, even if it failed the handshake.

jordan-rash (commented):
What happens if I start two nodes next to each other and the second one doesn't have "preserve network"?

internal/node/config.go
jordan-rash: meh, okay

kthomas:
> What happens if I start two nodes next to each other and the second one doesn't have "preserve network"?

Surely something bad would happen. You could accidentally do that now by starting a node and manually purging CNI. We need to handle moving to a new subnet when the first is fully utilized. We can consider this edge case at that time as well.

jordan-rash (approved):
Amazing

PR #119, opened 2024-02-12
Publish function exec pass and fail cloud events with log emit

autodidaddict:


kthomas (changes requested):
`triggerHandler` should be an instance method on `MachineManager` and should probably be named `handleTrigger` to align with naming convention followed across other handlers.

autodidaddict:
> `triggerHandler` should be an instance method on `MachineManager` and should probably be named `handleTrigger` to align with naming convention followed across other handlers.

Ok i'll make it an instance method. I go back and forth on naming conventions for funcs that return functions... I'll see what I can come up with.

kthomas:
There is one inconsistency in naming still between `publishFunctionExecSucceeded ` and `publishFuncExecFailed`. Otherwise LGTM.

kthomas (approved):

PR #122, opened 2024-02-19
chore: improve cache workload

yordis:
From what I can tell, my apologies if I miss something obvious,

There is no need to download the object into a file, then read the file back into memory, and delete the file afterward. We could use `store.GetBytes` instead.

### DIFF
--- a/internal/node/payload_cache.go
+++ b/internal/node/payload_cache.go
@@ -3,13 +3,9 @@ package nexnode
 import (
 	"crypto/sha256"
 	"encoding/hex"
-	"io"
 	"log/slog"
-	"os"
-	"path"
 	"strings"
 
-	"github.com/google/uuid"
 	"github.com/nats-io/nats.go"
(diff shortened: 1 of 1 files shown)


internal/node/payload_cache.go
```diff
@@ -56,26 +52,12 @@ func (m *MachineManager) CacheWorkload(request *controlapi.DeployRequest) (uint6
...
-		return 0, nil, err
-	}
-
-	workload, err := io.ReadAll(f)
-	if err != nil {
-		m.log.Error("Couldn't read the file we just wrote", slog.Any("err", err))
-		return 0, nil, err
-	}
-
-	os.Remove(filename)
```
yordis, line 77: Since we deleted the file right after doing `io.ReadAll`, we no longer use the file, as far as I can tell.

kthomas (approved):
LGTM!

PR #124, opened 2024-02-20
Increase rootfs size

kthomas:
This PR increases the default size of the agent rootfs image to 150 MiB.


autodidaddict (commented):
What is it that needs the default root size to increase? The smaller the rootfs the better, so I'm wondering what limit we're running into?

autodidaddict (approved):

PR #125, opened 2024-02-22
feat: add logger to wasm provider agent

yordis:


internal/agentlogger/agentlogger.go
```diff
@@ -0,0 +1,100 @@
...
+
+	subject := fmt.Sprintf("agentint.%s.logs", *a.md.VmID)
+	err = a.nc.Publish(subject, bytes)
+	if err != nil {
+		return err
+	}
+
+	// TODO: why is the logger controlling the flush?
+	//  is this to avoiding dealing with draining?
+	err = a.nc.Flush()
```
yordis, line 61: I need help here, any context about calling `Flush`? I am not sure why the logger would control that when the logger is capable to batch 64 entries in the channel

yordis:
The cyclical dependency is solvable, but I am not sure what the you would like to see happening,

- I remove the `agentlogger` from inside `agent-api`
- Remove `agent-api` from `agentlogger`
- Add an interface to `agent-api` and implement that interface in the `agentlogger`

jordan-rash:
Closing due to pending refactor of webassembly provider.  This will be addressed in next update

PR #126, opened 2024-02-23
Set VM metadata outside of VM creation

kthomas:
This commit handles some small cleanup items to make things more consistent.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -136,7 +136,7 @@ func (a *Agent) Start() {
 
 // Request a handshake with the host indicating the agent is "all the way" up
 // NOTE: the agent process will request a VM shutdown if this fails
-func (a *Agent) RequestHandshake() error {
+func (a *Agent) requestHandshake() error {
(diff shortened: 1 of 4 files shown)


autodidaddict:
Can you elaborate on what problems or situations this resolves? You mention that it makes things more consistent but I'm not sure how or how things were inconsistent prior to this.

autodidaddict:
Also need to make sure this passes the build/lint

kthomas:
Previously we were setting metadata within firecracker VM creation but it is more readable and extensible if that takes place in machine manager.

The inconsistencies were just related to syntax itself.

autodidaddict:
Ok I thought maybe there was a race condition involving _when_ we set the metadata, not just a change of roles as to which struct sets it. As long as the metadata is still being set at the right time, this looks fine.

kthomas:
> Also need to make sure this passes the build/lint

It passes.

autodidaddict (approved):

PR #133, opened 2024-03-01
Adds max bytes setting to ad hoc storage for devrun

autodidaddict:
In many situations, including NGS/Synadia Cloud, you can't create object store buckets without specifying a maximum number of bytes. To keep the `devrun` option the easiest way to get started, this adds a max bytes of **100MB** to the configuration.

The worst case scenario is eventually the developer has a workload that's > 100MB, and they have to update the bucket configuration.


kthomas (approved):
LGTM!

PR #136, opened 2024-03-04
Adds block to prevent deploying to un-handshake'd VM, shuts node down if first handshake fails

### LINKED ISSUE #134: Nex tries to deploy workload into a VM that hasn't reported a handshake
### Observed behavior

Start a nex node and have the agent within the first VM not report a handshake (this is pretty difficult to reproduce). A request to deploy a workload comes in and the node will attempt to deploy into that VM, and since the VM is effectively dead, the result will be a "nats: timeout"

### Expected behavior

The Nex node should abort and terminate if an agent fails to handshake. Agents failing to handshake is an indication of instability that is enough to warrant stopping the node. A missing handshake used to stop the node and no longer does, so this is a regression.

This should then never allow the situation where we attempt to deploy into a VM that hasn't given a handshake

### Nex and NATS version

Latest release of both

### Host environment

_No response_

### Steps to reproduce

_No response_

autodidaddict:
Closes #134 

Don't worry about the large "delete" portion of the diff, I just moved a bunch of related functions into their own file to try and keep the `machine_mgr.go` file to a manageable size.


internal/node/machine_mgr.go
```diff
@@ -344,6 +342,11 @@ func (m *MachineManager) awaitHandshake(vmid string) {
 	for !handshakeOk && !m.stopping() {
 		if time.Now().UTC().After(timeoutAt) {
 			m.log.Error("Did not receive NATS handshake from agent within timeout.", slog.String("vmid", vmid))
+			if len(m.handshakes) == 0 {
+				m.log.Error("First handshake failed, shutting down to avoid inconsistent behavior")
+				_ = m.Stop()
```
kthomas, line 347: You can call `m.ctx.Cancel()` here instead of `m.Stop()` and `os.Exit()`.

We need to add a call to `n.shutdown()` as line 124 of internal/node.go.

internal/node/machine_mgr.go
```diff
@@ -352,133 +355,6 @@ func (m *MachineManager) awaitHandshake(vmid string) {
 	}
 }
 
-// Called when the node server gets a log entry via internal NATS. Used to
```
kthomas, line 355: I kinda liked having all of the `MachineManager` code in the same place. Maybe it doesn't matter too much. @jordan-rash do you have an opinion?

internal/node/machine_mgr.go
kthomas: If we merge #137 you can then call `m.cancel()` instead of `m.Stop()`. This will result in the node context cleaning up everything properly 👍🏻

internal/node/machine_mgr.go
autodidaddict: I start to itch when we get over the 500 line mark. While yes it's all under the machine manager umbrella, I like having logically related stuff in nearby locations.

internal/node/machine_mgr.go
```diff
@@ -347,6 +345,11 @@ func (m *MachineManager) awaitHandshake(vmid string) {
 	for !handshakeOk && !m.stopping() {
 		if time.Now().UTC().After(timeoutAt) {
 			m.log.Error("Did not receive NATS handshake from agent within timeout.", slog.String("vmid", vmid))
+			if len(m.handshakes) == 0 {
+				m.log.Error("First handshake failed, shutting down to avoid inconsistent behavior")
+				m.cancel()
+				os.Exit(1)
```
kthomas, line 351: You can also drop this call to `os.Exit()` - the cancel function will result in a 0 exit code but the error will be logged.

internal/node/machine_mgr.go
kthomas: > I start to itch when we get over the 500 line mark. While yes it's all under the machine manager umbrella, I like having logically related stuff in nearby locations.

Makes sense. It would be better if we created a new interface and struct to make things more readable imho instead of separating it into separate files. I will not die on this hill though.

PR #138, opened 2024-03-05
Fix default slog initialization

kthomas:
The `--loglevel` option was no longer being applied.

### DIFF
--- a/nex/nex.go
+++ b/nex/nex.go
@@ -139,7 +139,7 @@ func main() {
 	if Opts.LogJSON {
 		logger = slog.New(slog.NewJSONHandler(os.Stdout, &opts))
 	} else {
-		logger = slog.Default()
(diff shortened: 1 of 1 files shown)


jordan-rash (changes requested):
@autodidaddict  wants the TextHandler to output the same as the Default logger....i haven't had time to look into this yet.  Thats why this was switched to Default

autodidaddict (approved):

kthomas:
@jordan-rash can we put this back for now and figure out a solution to make it pretty in a subsequent PR?

I don't disagree that the `TextWriter` log output is less than ideal, but in the meantime we need to be able to see things logs below `Info`.

jordan-rash:
Fo sho

jordan-rash (approved):

PR #144, opened 2024-03-19
ngs needs this value set to create streams

jordan-rash:


autodidaddict (changes requested):
~that's only 500kb for the binary. If we're going to hardcode a max for now, let's hard code it at a higher value, like 100mb~

autodidaddict (approved):
ignore me. I thought we were setting max bytes on an object store.

PR #148, opened 2024-03-25
adds name to nats connection

jordan-rash:
Names nats connection based on if its a node or a cli command

### DIFF
--- a/internal/models/cli.go
+++ b/internal/models/cli.go
@@ -147,7 +147,7 @@ type TokenBucket struct {
 	Size *int64 `json:"size"`
 }
 
-func GenerateConnectionFromOpts(opts *Options, connectionName string) (*nats.Conn, error) {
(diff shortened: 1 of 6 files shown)


internal/models/cli.go
```diff
@@ -147,7 +147,7 @@ type TokenBucket struct {
 	Size *int64 `json:"size"`
 }
 
-func GenerateConnectionFromOpts(opts *Options) (*nats.Conn, error) {
+func GenerateConnectionFromOpts(opts *Options, connectionName string) (*nats.Conn, error) {
```
kthomas, line 150: There is a `ConnectionName` in `Options`...

jordan-rash: dang....thanks 😅

kthomas (approved):

PR #149, opened 2024-03-26
add flag for nats connection name

jordan-rash:


nex/config.json
```diff
@@ -0,0 +1,15 @@
+{
```
kthomas, line 1: Did you mean to commit this?

nex/nex.go
```diff
@@ -71,6 +71,7 @@ func init() {
 	ncli.Flag("logjson", "Log JSON").Default("false").Envar("NEX_LOGJSON").UnNegatableBoolVar(&Opts.LogJSON)
 	ncli.Flag("context", "Configuration context").Envar("NATS_CONTEXT").PlaceHolder("NAME").StringVar(&Opts.ConfigurationContext)
 	ncli.Flag("no-context", "Disable NATS context discovery").UnNegatableBoolVar(&Opts.SkipContexts)
+	ncli.Flag("conn-name", "Name of NATS connection").StringVar(&Opts.ConnectionName)
```
autodidaddict, line 74: If this flag is left off, should we use a reasonable default, like `nex-vXX` so the default behavior is to identify nex and its version?

nex/config.json
autodidaddict, line 1: Are we planning on checking this file in, or is this an artifact of running nex?

autodidaddict (changes requested):

nex/config.json
jordan-rash: YESSSSS!!!!....no

kthomas (approved):

nex/nex.go
jordan-rash: done!

PR #154, opened 2024-03-27
Fix/refactormgrs

autodidaddict:
When testing this locally, the v8 runner is rejecting the echofunction.js example with a reason of `[]`. I don't know why this is happening and don't know if it's related to this PR or something else.

Elf binaries appear to work just fine.


autodidaddict:
Now verified that this does indeed work with both elf and v8 binaries

internal/agent-api/client.go
```diff
@@ -0,0 +1,191 @@
...
+			slog.String("agentId", a.agentId), slog.String("error", err.Error()))
+		return err
+	}
+	return nil
+}
+
+func (a *AgentClient) awaitHandshake(agentId string) {
+	// timeoutAt := time.Now().UTC().Add(a.handshakeTimeout)
+
+	<-time.After(a.handshakeTimeout)
```
kthomas, line 127: This pattern will prevent the node from stopping in a timely manner.

The previous machine manager code allowed the node to exit without waiting for this timeout:

```
func (m *MachineManager) awaitHandshake(vmid string) {
	timeoutAt := time.Now().UTC().Add(m.handshakeTimeout)

	handshakeOk := false
	for !handshakeOk && !m.stopping() {
		if time.Now().UTC().After(timeoutAt) {
			m.log.Error("Did not receive NATS handshake from agent within timeout.", slog.String("vmid", vmid))
			if len(m.handshakes) == 0 {
				m.log.Error("First handshake failed, shutting down to avoid inconsistent behavior")
				m.cancel()
			}
			return
		}

		_, handshakeOk = m.handshakes[vmid]
		time.Sleep(time.Millisecond * agentapi.DefaultRunloopSleepTimeoutMillis)
	}
}
```

internal/node/machine_mgr_events.go
```diff
@@ -48,13 +47,13 @@ func (m *MachineManager) publishFunctionExecSucceeded(vm *runningFirecracker, ts
...
-	subject := fmt.Sprintf("%s.%s.%s.%s.%s", LogSubjectPrefix, vm.namespace, m.publicKey, *vm.deployRequest.WorkloadName, vm.vmmID)
-	err = m.nc.Publish(subject, logBytes)
+	subject := fmt.Sprintf("%s.%s.%s.%s.%s", LogSubjectPrefix, vm.namespace, m.node.publicKey, *vm.deployRequest.WorkloadName, vm.vmmID)
+	err = m.node.nc.Publish(subject, logBytes)
 	if err != nil {
 		m.log.Error("Failed to publish function exec passed log", slog.Any("err", err))
 	}
 
-	return m.nc.Flush()
+	return m.node.nc.Flush()
```
kthomas, line 56: The machine manager should not need to know about the node.

internal/agent-api/client.go
autodidaddict: the timeout is a tiny number. The above code didn't actually work anymore. I was constantly getting handshake failures. It didn't start working again until I used the new method.

Waiting the remainder of a timeout measured in milliseconds before a node shuts down isn't a bad thing

kthomas (changes requested):
The encapsulation provided by the `Node` -> `MachineManager` was to prevent us from breaking law of demeter. All of the new references now to `m.node.*` should be reconsidered.

autodidaddict:
> The encapsulation provided by the `Node` -> `MachineManager` was to prevent us from breaking law of demeter. All of the new references now to `m.node.*` should be reconsidered.

yeah the whole point of me doing small refactoring PRs was to try and get incremental stuff done. Machine Manager in its current form is going away, in favor a `WorkloadManager`, which will encapsulate the workload management strategy and ultimately support both firecracker and firecrackerless process managers

/shrug

internal/node/machine_mgr_events.go
```diff
@@ -150,31 +149,57 @@ func (m *MachineManager) publishMachineStopped(vm *runningFirecracker) error {
...
+// 	if vm.deployRequest != nil {
+// 		workload = vm.deployRequest.WorkloadName
+// 	}
+
+// 	subject := logPublishSubject(vm.namespace, m.node.publicKey, vmID, workload)
+// 	_ = m.nc.Publish(subject, bytes)
+// }
+
+func (m *MachineManager) agentLog(agentId string, entry agentapi.LogEntry) {
+	vm, ok := m.allVMs[agentId]
```
kthomas, line 193: The map of VMs should still be indexed on vm id. Did we just wholesale rename vm id to agent id?

Instead, we should probably index these based on workload id. Which should come from the `WorkloadManager` eventually...

internal/node/machine_mgr_events.go
```diff
@@ -150,31 +149,57 @@ func (m *MachineManager) publishMachineStopped(vm *runningFirecracker) error {
...
-
-	m.log.Debug("Received agent log", slog.String("vmid", vmID), slog.String("log", logentry.Text))
-
 	bytes, err := json.Marshal(&emittedLog{
-		Text:      logentry.Text,
-		Level:     slog.Level(logentry.Level),
-		MachineId: vmID,
+		Text:      entry.Text,
+		Level:     slog.Level(entry.Level),
+		MachineId: agentId,
```
kthomas, line 202: `MachineId = agentId`? My brain hurts.

internal/node/process_mgr.go
```diff
@@ -0,0 +1,17 @@
+package nexnode
+
+import agentapi "github.com/synadia-io/nex/internal/agent-api"
+
+// TODO: this will be utilized in a forthcoming PR
+type ProcessManager interface {
```
kthomas, line 6: Is this meant to be what I was calling `WorkloadManager`?

kthomas: I strongly suggest we call this `WorkloadManager` -- it does indeed seem to be that. I like this interface very much.

spec/node_test.go
```diff
@@ -384,32 +384,6 @@ var _ = Describe("nex node", func() {
 							Expect(managerProxy.Telemetry()).To(Equal(nodeProxy.Telemetry()))
 						})
 
-						Describe("agent internal API subscriptions", func() {
```
kthomas, line 387: These should all still be valid tests.

internal/node/machine_mgr.go
```diff
@@ -338,53 +319,17 @@ func (m *MachineManager) LookupMachine(vmId string) *runningFirecracker {
...
-			if len(m.handshakes) == 0 {
-				m.log.Error("First handshake failed, shutting down to avoid inconsistent behavior")
-				m.cancel()
-			}
-			return
-		}
-
-		_, handshakeOk = m.handshakes[vmid]
-		time.Sleep(time.Millisecond * agentapi.DefaultRunloopSleepTimeoutMillis)
+func (m *MachineManager) agentHandshakeTimedOut(agentId string) {
```
kthomas, line 322: Does this get called anywhere else other than `awaitHandshake`? If not, it makes it harder to read as a separate function.

What will the use-cases be to allow `AgentClient` instances to support this provided timeout callback?

kthomas:
> > The encapsulation provided by the `Node` -> `MachineManager` was to prevent us from breaking law of demeter. All of the new references now to `m.node.*` should be reconsidered.
> 
> yeah the whole point of me doing small refactoring PRs was to try and get incremental stuff done. Machine Manager in its current form is going away, in favor a `WorkloadManager`, which will encapsulate the workload management strategy and ultimately support both firecracker and firecrackerless process managers
> 
> /shrug

I actually think `WorkloadManager` should hold on to `MachineManager` in nearly its current form. Then we would be able to define a `Manager` interface, to be implemented by both `MachineManager` and `AgentManager`. This would allow us to do even smaller refactoring by keeping working code working, and simply tucking `MachineManager` behind a new `WorkloadManager` type.

autodidaddict:
That's pretty nearly what I plan. Review still says you're waiting for changes

kthomas (approved):

PR #156, opened 2024-04-02
allow agent to pull its metadata from environment

autodidaddict:
In order to eventually support the no-sandbox mode, agents will need a way to get the NATS internal IP and port and the workloadId that it's been assigned from the node.

This PR makes it so that if the `NEX_SANDBOX` environment variable both exists and contains the value "false", then the agent will skip the MMDS query. When the agent is running in a sandbox (firecracker + rootfs), the agent is running as an openrc service and so won't have that env var.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"path"
+	"strings"
 	"sync/atomic"
 	"syscall"
 	"time"
@@ -61,8 +62,8 @@ func NewAgent(ctx context.Context, cancelF context.CancelFunc) (*Agent, error) {
 	var metadata *agentapi.MachineMetadata
 	var err error
 
-	if os.Getenv(nexEnvSandbox) == "false" {
-		metadata, err = GetMachineDataFromEnv()
+	if strings.ToLower(os.Getenv(nexEnvSandbox)) == "false" {
+		metadata, err = GetMachineMetadataFromEnv()
 	} else {
 		metadata, err = GetMachineMetadata()
 	}
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -58,14 +58,21 @@ func HaltVM(err error) {
 
 // Initialize a new agent to facilitate communications with the host
 func NewAgent(ctx context.Context, cancelF context.CancelFunc) (*Agent, error) {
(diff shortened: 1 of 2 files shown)


agent/agent.go
```diff
@@ -58,14 +58,21 @@ func HaltVM(err error) {
 
 // Initialize a new agent to facilitate communications with the host
 func NewAgent(ctx context.Context, cancelF context.CancelFunc) (*Agent, error) {
-	metadata, err := GetMachineMetadata()
+	var metadata *agentapi.MachineMetadata
+	var err error
+
+	if os.Getenv(nexEnvSandbox) == "false" {
```
kthomas, line 64: Might want to do a `strings.ToLower` here.

agent/agent.go
```diff
@@ -58,14 +58,21 @@ func HaltVM(err error) {
...
+		metadata, err = GetMachineDataFromEnv()
```
kthomas, line 65: Calling it "data" here and "metadata" below. Should probably call this `GetMachineMetadataFromEnv()` so it matches `GetMachineMetadata()`.

autodidaddict: agreed. VScode's automatic "suggestion" here burned me.

agent/metadata.go
```diff
@@ -15,6 +17,11 @@ import (
 // (see https://github.com/firecracker-microvm/firecracker/blob/main/docs/mmds/mmds-user-guide.md#version-2)
 const MmdsAddress = "169.254.169.254"
 
+const nexEnvSandbox = "NEX_SANDBOX"
+const nexEnvVmID = "NEX_VMID"
+const nexEnvNodeNatsHost = "NEX_NODENATSHOST"
```
kthomas, line 22: Might want to use underscores so it's more human-readable. It isn't a big deal now but we could end up having a lot more of these env vars later. `NEX_NODE_NATS_HOST` is a lot easier on the eyes.

kthomas (approved):
I left a few suggestions but LGTM otherwise. We can clean those items up later if you prefer.

agent/metadata.go
autodidaddict: good call

autodidaddict:
> I left a few suggestions but LGTM otherwise. We can clean those items up later if you prefer.

Addressed your items

PR #158, opened 2024-04-04
adds the spawning process manager to run with scissors

autodidaddict:


internal/node/spawn_procman.go
```diff
@@ -33,50 +64,215 @@ func NewSpawningProcessManager(
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+type procLogEmitter struct {
+	stderr bool
+	// TODO: personal opinion - not sure I like propagating logger instances everywhere...
```
kthomas, line 259: Agree, I don't think it would be the end of the world to reference a global `Log` instance that was setup on package `init`.

agent/agent.go
```diff
@@ -34,26 +35,32 @@ type Agent struct {
...
-	err = syscall.Reboot(syscall.LINUX_REBOOT_CMD_RESTART)
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "Failed to reboot: %s", err)
+	if isSandboxed() {
+		err = syscall.Reboot(syscall.LINUX_REBOOT_CMD_RESTART)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Failed to halt: %s", err)
+		}
+	} else {
+		os.Exit(code)
```
kthomas, line 63: I am not a huge fan of calling `os.Exit` as it prevents graceful cleanup. Still reviewing this PR though so will hold any suggestions on how we might do it differently for later.

agent/agent.go
```diff
@@ -34,26 +35,32 @@ type Agent struct {
 	cancelF context.CancelFunc
 	closing uint32
 	ctx     context.Context
-	sigs    chan os.Signal
```
kthomas, line 37: Why don't we want to handle signals in the `Agent`? It would seemingly make even _more_ sense to support signal handling in no-sandbox mode.

agent/agent.go
```diff
@@ -398,15 +408,42 @@ func (a *Agent) newExecutionProviderParams(req *agentapi.DeployRequest, tmpFile
 	return params, nil
 }
 
+func (a *Agent) setupSignalHandlers() {
```
kthomas, line 411: Similar concern regarding use of `os.Exit` here in `setupSignalHandlers`.

agent/agent.go
```diff
@@ -398,15 +408,42 @@ func (a *Agent) newExecutionProviderParams(req *agentapi.DeployRequest, tmpFile
...
+				os.Exit(0)
+			}
+		}
+	}()
+}
+
 func (a *Agent) shutdown() {
 	if atomic.AddUint32(&a.closing, 1) == 1 {
+		if a.provider != nil {
+			err := a.provider.Undeploy()
```
kthomas, line 434: This is going to probably result in `Undeploy()` being called twice when a workload has been deployed, which would require an additional mutex to ensure that call is idempotent.

agent/agent.go
autodidaddict: We are handling signals. We weren't before. I'm just not using a channel to do it

agent/agent.go
autodidaddict: this os.Exit is actually called inside the graceful cleanup handler

agent/agent.go
autodidaddict: I can't see how this will call undeploy twice.

agent/agent.go
```diff
@@ -68,11 +75,12 @@ func NewAgent(ctx context.Context, cancelF context.CancelFunc) (*Agent, error) {
 		metadata, err = GetMachineMetadata()
 	}
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "failed to get machine metadata: %s", err)
-		return nil, err
+		fmt.Fprintf(os.Stderr, "failed to get machien metadata: %s\n", err)
```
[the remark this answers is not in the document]
autodidaddict: until this PR, the agent had no ability to emit text logs at all, it was all done through the channels which in turn published NATS events. We can upgrade the logging here in another PR

internal/node/spawn_procman.go
jordan-rash: oof, its not often I am on the opposite side of the line!  I would really not use globals at all if possible.  You instantly loose context and don't gain anything in return

agent/agent.go
jordan-rash: FWIW, os.Exit doesn't allow any pending `defer` to execute.  we should probably be using `runtime.Goexit` https://pkg.go.dev/runtime#Goexit

internal/node/spawn_procman.go
kthomas: Logging is about the only thing that I will compromise on. I don't mind passing it around though, for the reasons you stated 😏

agent/agent.go
jordan-rash: https://pkg.go.dev/runtime#Goexit

agent/agent.go
```diff
@@ -448,3 +485,7 @@ func (a *Agent) workAck(m *nats.Msg, accepted bool, msg string) error {
 
 	return nil
 }
+
+func isSandboxed() bool {
+	return strings.ToLower(os.Getenv("NEX_SANDBOX")) != "false"
```
jordan-rash, line 490: what happens if we want to run no-sandbox on a node we can't go set the environment on?  Should the `nex [dev]run` command be telling the agent that it is sandboxed or not?

examples/nodeconfigs/nosandbox.json
```diff
@@ -0,0 +1,17 @@
...
+        "interface_name": "veth0"
+    },
+    "machine_template": {
+        "vcpu_count": 1,
+        "memsize_mib": 256
+    },
+    "tags": {
+        "simple": "true"
+    },
+    "no_sandbox": true
```
jordan-rash, line 16: personal preference, double negates confuse me. 
sandbox: [true|false] would be more clear i think

internal/node/spawn_procman.go
```diff
@@ -33,50 +64,215 @@ func NewSpawningProcessManager(
...
+}
+
+// Spawns a new child process, a waiting nex-agent
+func (s *SpawningProcessManager) spawn() (*spawnedProcess, error) {
+	id := xid.New()
+	workloadID := id.String()
+
+	cmd := exec.Command(nexAgentBinary)
+	cmd.Env = append(os.Environ(),
+		"NEX_SANDBOX=false",
```
jordan-rash, line 214: i retract my previous question

jordan-rash (approved):
This all looks good to me, only nits.  That said, I still need to put all the changes from 155 apart to understand the full flow of everthing now, just havent had time yet

agent/agent.go
kthomas: Not sure we need `Goexit` here either, although this is interesting.

agent/agent.go
kthomas: I believe that in at least some cases it will get called once when `agentint.<id>.undeploy` is handled and a second time when the agent is shutting down.

autodidaddict: fixed

agent/agent.go
autodidaddict: the run command should never be able to dictate whether the node is sandboxing or not, mainly for security reasons. We can add a tag to the node indicating that it's unsafe

examples/nodeconfigs/nosandbox.json
autodidaddict: I actually used the double negative like this deliberately so people couldn't accidentally disable the sandbox

Issue #61, opened 2024-01-18
Change uptime to runtime and modify calculation to support functions


autodidaddict:
Current implementation uses the difference between now and the original deploy/launch time to determine the `uptime` value for a given workload. This value has the potential to confuse people. By changing to `runtime`, that value should now have the same meaning for both functions and services/elf, though calculated differently.

* For services - the value will continue to be calculated as it is currently: `now - deploy time` 
* For functions - the value is an accumulated sum of actual execution time. We will time each execution of the function and add that time to a growing counter that will be used for the total elapsed execution time (runtime). 

Having runtime be an accumulator for functions also makes for a nice prometheus metric.

We should also continue to maintain the original deployment time and make that available to control client queries, as it might be really useful to know that a function was deployed 3 weeks ago and has only accrued 12 minutes of total execution time.

autodidaddict:
Closed by #166

PR #166, opened 2024-04-12
makes uptimes and runtimes real

autodidaddict:


autodidaddict:
Closes #61

internal/node/workload_mgr.go
```diff
@@ -293,15 +293,39 @@ func (w *WorkloadManager) RunningWorkloads() ([]controlapi.MachineSummary, error
 	summaries := make([]controlapi.MachineSummary, len(procs))
 
 	for i, p := range procs {
+		uptimeFriendly := "unknown"
```
kthomas, line 296: Might be good to break this out into an `enrich()` method... someday.

internal/agent-api/client.go
```diff
@@ -162,6 +166,19 @@ func (a *AgentClient) Undeploy() error {
 	return nil
 }
 
+func (a *AgentClient) RecordExecTime(elapsedNanos int64) {
+	a.execTotalNanos += elapsedNanos
```
kthomas, line 170: This isn't thread-safe. If a function is called at the same exact moment, there will be a race to set this.

kthomas: Could simply add an execTimeMutex in the agent and lock/defer unlock before `a.execTotalNanos += elapsedNanos`

autodidaddict: if this codebase gets one more mutex, I'm rewriting all the async code ;)

kthomas: Yeah, `atomic` is a better way to do it ;)

kthomas (approved):

internal/node/workload_mgr.go
autodidaddict: there's already `myUptime`, but since I copied that from NATS server I didn't want to mess with it... yet.

PR #172, opened 2024-04-15
Add v8 http host service specs

kthomas:
This PR adds initial specs for the v8 http host service.

### DIFF
--- a/internal/node/workload_mgr_events.go
+++ b/internal/node/workload_mgr_events.go
@@ -116,12 +116,6 @@ func (w *WorkloadManager) publishFunctionExecFailed(workloadId string, workload 
 		return nil
 	}
 
-	// FIXME-- namespace should always be non-nil here...
-	namespace := "default"
-	if deployRequest.Namespace != nil {
-		namespace = *deployRequest.Namespace
-	}
-
 	functionExecFailed := struct {
 		Name      string `json:"workload_name"`
 		Subject   string `json:"trigger_subject"`
@@ -129,7 +123,7 @@ func (w *WorkloadManager) publishFunctionExecFailed(workloadId string, workload 
 		Error     string `json:"error"`
 	}{
 		Name:      workload,
-		Namespace: namespace,
+		Namespace: *deployRequest.Namespace,
 		Subject:   tsub,
 		Error:     origErr.Error(),
 	}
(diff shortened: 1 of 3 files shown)


autodidaddict (dismissed):

internal/node/workload_mgr_events.go
```diff
@@ -116,14 +116,20 @@ func (w *WorkloadManager) publishFunctionExecFailed(workloadId string, workload
 		return nil
 	}
 
+	// FIXME-- namespace should always be non-nil here...
+	namespace := "default"
```
jordan-rash, line 120: does the namespace not come from the CLI arguments?  meaning, we should have a default set, right? 
https://github.com/synadia-io/nex/blob/f644d1d58692b03cf28adb254610400c767755bd/nex/nex.go#L71

autodidaddict: correct, it's always populated with _something_

kthomas: The FIXME is indeed a hint for us to audit the node API and ensure we validate client input, as suggested by @autodidaddict.

I can investigate where we are missing that check internal to the node API and add a commit to this PR to address this.

autodidaddict: we aren't missing that check. The NATS subscription does this for us:

https://github.com/synadia-io/nex/blob/main/internal/node/controlapi.go#L83

we literally can't receive a deploy message unless it contains at least a 1-character namespace

kthomas: True... strange that I was seeing a nil pointer exception without the check there. Will figure out why that is shortly and amend this PR.

kthomas: The tests are using this method:

```
nodeClient := controlapi.NewApiClientWithNamespace(_fixtures.natsConn, time.Millisecond*1000, "default", log)
_, err = nodeClient.StartWorkload(deployRequest)
```

The tests are also manually constructing a controlapi deploy request, which is what is passed into the above `StartWorkload` call. Looking at how to resolve this and remove the FIXME.

kthomas: This can be re-reviewed. Reverted the unnecessary commit and FIXME 👍🏻

autodidaddict (approved):

PR #177, opened 2024-04-16
Adds lame duck mode

autodidaddict:


internal/node/node.go
```diff
@@ -135,6 +136,23 @@ func (n *Node) Stop() {
 	n.shutdown()
 }
 
+func (n *Node) EnterLameDuck() error {
+	atomic.StoreUint32(&n.lameduck, 1)
```
kthomas, line 140: Should there be a check here to make sure `n.lameduck` isn't already 1 to prevent accidentally calling it twice?

kthomas: Currently it seems `procMan.EnterLameDuck()` is idempotent for both of the implementations but it seems like a bad assumption that they will always behave that way.

autodidaddict: good call. made it only happen once.

kthomas (dismissed):

kthomas:

jordan-rash (approved):

PR #180, opened 2024-04-17
Add initial object store host service for v8 workloads

kthomas:


internal/node/services/lib/objectstore.go
```diff
@@ -43,6 +51,14 @@ func (o *ObjectStoreService) HandleRPC(msg *nats.Msg) {
 	method := tokens[6]
 
 	switch method {
+	case objectStoreServiceMethodGet:
```
autodidaddict, line 54: tiny nit... we have the split token slice already, and all of these handle methods are calling `split` again. Would be (slightly) more efficient to just pass the one split slice in as a parameter so the handlers don't need to re-split.

autodidaddict (approved):

internal/node/services/lib/objectstore.go
kthomas: Can address this as part of cleanup across all host services 👍🏻

PR #182, opened 2024-04-17
Provide OTel solution for local development

jordan-rash:


README.md
```diff
@@ -41,7 +41,9 @@ The pre-flight check requires a configuration file so it can perform the appropr
     },
     "tags": {
         "simple": "true"
-    }
+    },
+    "no_sandbox": false,
```
autodidaddict, line 45: we don't need to specify this if the value is false. It will default to using the sandbox unless we override it.

autodidaddict (approved):

autodidaddict:
Having trouble seeing the CLI support behavior. if you specify the trace exporter via command line, does that override what's inside the config file?

jordan-rash:
> Having trouble seeing the CLI support behavior. if you specify the trace exporter via command line, does that override what's inside the config file?

Thats because your too fast for me.  Check now

autodidaddict:
> > Having trouble seeing the CLI support behavior. if you specify the trace exporter via command line, does that override what's inside the config file?
> 
> Thats because your too fast for me. Check now

I still can't find where the CLI-supplied values override the config-file value :eyes:

jordan-rash:
> > > Having trouble seeing the CLI support behavior. if you specify the trace exporter via command line, does that override what's inside the config file?
> > 
> > 
> > Thats because your too fast for me. Check now
> 
> I still can't find where the CLI-supplied values override the config-file value 👀 From the code I'm looking at, it looks like it only uses the configuration file value and the CLI values are ignored

Oh, the CLI doesnt overwrite the `otlp_exporter_url` option.  All thats added is a bool to actually use trace (irregardless of the state of `otlp_exporter_url`) and which exporter to use (which at the moment, there is only one -> grpc).

kthomas:

kthomas (approved):

PR #183, opened 2024-04-17
Makes CNI config more flexible

jordan-rash:


autodidaddict (approved):

internal/node/config.go
```diff
@@ -22,6 +22,11 @@ func LoadNodeConfiguration(configFilepath string) (*models.NodeConfiguration, er
 		return nil, err
 	}
 
+	// TODO: should we validate the configuration here
```
kthomas, line 25: I don't think so. We should be validating it in `NewNode`, and it returns an error if that validation fails.

See https://github.com/synadia-io/nex/blob/main/internal/node/node.go#L75

jordan-rash: maybe both?  why wouldnt we validate it right after its created?

kthomas: It wouldn't hurt anything if we did it inside of `LoadNodeConfiguration` sure -- but if so, do it at the very end before you return the config.

kthomas:

internal/node/config.go
jordan-rash: 👍🏼

autodidaddict (approved):

kthomas (approved):

PR #185, opened 2024-04-17
Publish heartbeats on the main node goroutine

kthomas:
I didn't recognize this when I first reviewed the PR, but we can publish heartbeats with one less goroutine.


autodidaddict (approved):
There was no need for this refactor.. there wasn't an issue with having that one Go routine and I find this version harder to read... but personal preference is what it is.

PR #187, opened 2024-04-18
Ping can now optionally filter by workload ID or name

autodidaddict:
You can now use a different `WPING` endpoint to have nodes _only_ reply to a ping if they're currently running the indicated workload, which can be the workload ID or the workload name.

So if you specify a workload ID, you should expect 0 or 1 nodes.
If you specify a workload name, you might get multiple nodes if that workload is scaled out.

Added the `--workload` option to the `node ls` command, so the end user doesn't need to know that this is a different ping subject under the hood.

### DIFF
--- a/control-api/client.go
+++ b/control-api/client.go
@@ -138,7 +138,7 @@ func (api *Client) ListNodesWithWorkload(workloadId string) ([]PingResponse, err
 		responses = append(responses, resp)
 	})
 	if err != nil {
-		return nil, nil
+		return nil, err
 	}
 	msg := nats.NewMsg(fmt.Sprintf("%s.WPING.%s.%s", APIPrefix, api.namespace, workloadId))
 	msg.Reply = sub.Subject
--- a/control-api/client.go
+++ b/control-api/client.go
@@ -112,6 +112,45 @@ func (api *Client) EnterLameDuck(nodeId string) (*LameDuckResponse, error) {
 	return &response, nil
 }
 
+// Functions the same as ListNodes (ping), except that only nodes currently running
+// the workload in question will respond. Useful to answer the questions, "is this workload running?"
+// and if so, on which node?
+func (api *Client) ListNodesWithWorkload(workloadId string) ([]PingResponse, error) {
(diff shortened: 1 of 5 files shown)


autodidaddict:
I had a variant of this where we reused the same `PING` endpoint and the optional workload filter was in the payload. I ended up not liking how this turned out because the regular pings actually return errors to the client, and the workload ping is silent on error...

control-api/client.go
```diff
@@ -112,6 +112,45 @@ func (api *Client) EnterLameDuck(nodeId string) (*LameDuckResponse, error) {
...
+			return
+		}
+		err = json.Unmarshal(bytes, &resp)
+		if err != nil {
+			return
+		}
+		responses = append(responses, resp)
+	})
+	if err != nil {
+		return nil, nil
```
jordan-rash, line 141: Return `nil, err`?

nex/nex.go
```diff
@@ -122,6 +122,9 @@ func init() {
 	rootfs.Flag("image", "Base image for rootfs build").Default("synadia/nex-rootfs:alpine").StringVar(&RootfsOpts.BaseImage)
 	rootfs.Flag("agent", "Path to agent binary").PlaceHolder("../path/to/nex-agent").Required().StringVar(&RootfsOpts.AgentBinaryPath)
 	rootfs.Flag("size", "Size of rootfs filesystem").Default(strconv.Itoa(1024 * 1024 * 150)).IntVar(&RootfsOpts.RootFSSize) // 150MB default
+
+	// one day when we refactor, let's get rid of all of these global structs. Such ugly
+	nodesLs.Flag("workload", "Only query nodes currently running the given workload (id or name)").StringVar(&RunOpts.Name)
```
jordan-rash, line 127: Maybe add a placeholder with a workloadID?

jordan-rash (dismissed):
Thanks Kevin!

nex/nex.go
autodidaddict: if I do that it'll just look like a nuid. and you can use either a nuid or the workload name... Given that it felt like I shouldn't have a placeholder

PR #190, opened 2024-04-19
Discussion: Test/spec module

jordan-rash:
Looking for thoughts on this...


autodidaddict:
Whole technically I think go will rip out unused deps in release builds, I am definitely in favor of  keeping the main app deps as slim as possible. 

Makes me wish for package managers that separate test and app dependencies explicitly

jordan-rash:
> While technically I think go will rip out unused deps in release builds, I am definitely in favor of keeping the main app deps as slim as possible.
> 
> Makes me wish for package managers that separate test and app dependencies explicitly

I believe you are correct, go does do this...however, all these modules are downloads and installed as part of the development and CI process (some big ones in there).  If you go peek at the Actions output, this is taking ~2 minutes off the build process.

autodidaddict:
> > While technically I think go will rip out unused deps in release builds, I am definitely in favor of keeping the main app deps as slim as possible.
> > 
> > Makes me wish for package managers that separate test and app dependencies explicitly
> 
> I believe you are correct, go does do this...however, all these modules are downloads and installed as part of the development and CI process (some big ones in there).  If you go peek at the Actions output, this is taking ~2 minutes off the build process.

Oh we're definitely on the same page here.

PR #194, opened 2024-04-23
support getting a list of workloads that match a ping

autodidaddict:
`$NEX.WPING.{namespace}` - probes all workloads in this namespace
`$NEX.WPING.{namespace}.{workload}` - probes all workloads in this namespace that match the supplied workload name/ID

the Nex control API client and CLI have been updated to reflect this.


autodidaddict:
This closes #194

internal/node/controlapi_test.go
```diff
@@ -0,0 +1,81 @@
+package nexnode
+
+import (
+	"slices"
+	"testing"
+
+	controlapi "github.com/synadia-io/nex/control-api"
+)
+
+func TestSummarizeMachinesForPing(t *testing.T) {
```
kthomas, line 10: Do we not want to keep these tests packaged alongside the others?

autodidaddict: this is a unit test. It is exactly where a unit test belongs in Go, in a `_test.go` file sitting next to the file under test.

kthomas: Other unit tests are under the `test/` package right now. As of the current state of how things are being run, this test file is not being included in the suite when it runs.

control-api/client.go
```diff
@@ -112,21 +112,25 @@ func (api *Client) EnterLameDuck(nodeId string) (*LameDuckResponse, error) {
 	return &response, nil
 }
 
-// Functions the same as ListNodes (ping), except that only nodes currently running
-// the workload in question will respond. Useful to answer the questions, "is this workload running?"
-// and if so, on which node?
-func (api *Client) ListNodesWithWorkload(workloadId string) ([]PingResponse, error) {
+// This is a filtered node ping that returns only matching workloads.
+// A workloadId of "" will not filter by workload, and only
```
kthomas, line 116: `nil` (e.g., `workloadID *string`) would probably be better to read and document.

autodidaddict: I disagree. I've lost all my patience with the `*string` pattern. Empty strings are just as easy to read here and half as much of a PITA

internal/node/controlapi_test.go
autodidaddict: unit tests shouldn't be in `test/`. That's not the standard Go pattern. We run `go test -v ./...` in a root directory to fire off unit tests. Anything in the `test/` directory is acceptance/integration/etc and launched separately.

kthomas: I am not saying having unit tests live alongside their counterparts isn't correct I am just saying I am pretty certain we are not running this new test you just wrote.

kthomas: You need to modify how we are running the tests.

It is currently `go test -v -race ./test` in CI.

autodidaddict: that's how it runs the spec suite in `/test`. The unit tests are an entirely different concern and are usually run separately.

kthomas: This is how it runs the unit tests... not the spec suite...

kthomas (approved):

Issue #170, opened 2024-04-12
Formalize host services SDK


autodidaddict:
### Proposed change

Host services is the set of functionality provided to Nex workloads as a built-in convenience. For example, JavaScript functions get access to a key-value bucket, an object store, basic pub/sub, and an HTTP client.

This would extract host services client code into its own "package"

### Use case

This would let code in long-running services utilize host services if they wanted to, giving people more built-in goodness when they use Nex deployments.

jordan-rash:
@kthomas should this be closed by #195??

autodidaddict:
This is a broader scope than the cleanup. For example, the host services client isn't a thing that can be used by external code yet

kthomas:
> @kthomas should this be closed by #195??

#195 was needed as a prereq to formalizing an SDK, in that message headers and non-enveloped payloads were made consistent across the v8 implementation.

autodidaddict:
Working on this now

PR #195, opened 2024-04-23
Cleanup host services

kthomas:
This PR cleans up the v8 host services provider implementation.

The agent v8 provider now exclusively requires and uses `Uint8Array` when:
- marshaling parameters included in functions called from javascript (`v8.Value` -> `[]byte`)
- unmarshaling raw bytes returned to functions called from javascript (`[]byte` -> `v8.Value`)

Additionally, the host services API has been cleaned up in preparation for formalizing an SDK (#170). Namely, the use of message headers and non-enveloped payloads have been made consistent across host services.

### DIFF
--- a/agent/providers/lib/v8.go
+++ b/agent/providers/lib/v8.go
@@ -63,11 +63,12 @@ const (
 	hostServicesObjectStoreDeleteTimeout = time.Millisecond * 3000
 	hostServicesObjectStoreListTimeout   = time.Millisecond * 3000
 
-	v8FunctionArrayAppend       = "array-append"
-	v8FunctionArrayInit         = "array-init"
-	v8FunctionUInt8ArrayInit    = "uint8-array-init"
-	v8FunctionUInt8ArraySetIdx  = "uint8-array-set-idx"
-	v8FunctionUInt8ArrayToArray = "uint8-array-to-array"
+	v8FunctionArrayAppend        = "array-append"
+	v8FunctionArrayInit          = "array-init"
+	v8FunctionUInt8ArrayInit     = "uint8-array-init"
+	v8FunctionUInt8ArraySetIdx   = "uint8-array-set-idx"
+	v8FunctionUInt8ArrayToArray  = "uint8-array-to-array"
+	v8FunctionUInt8ArrayToString = "uint8-array-to-bytes"
 
 	v8ExecutionTimeoutMillis = 5000
 	v8MaxFileSizeBytes       = int64(12288) // arbitrarily ~12K, for now
@@ -185,6 +186,8 @@ func (v *V8) Execute(subject string, payload []byte) ([]byte, error) {
 
 	select {
 	case val := <-vals:
+		// FIXME-- switch on val type or are we ok with forcing a JSON response?
+
 		retval, err := val.MarshalJSON()
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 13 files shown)


agent/providers/lib/v8.go
```diff
@@ -66,6 +66,10 @@ const (
 	nexTriggerSubject = "x-nex-trigger-subject"
 	nexRuntimeNs      = "x-nex-runtime-ns"
 
+	httpURL = "x-http-url"
+
+	keyHeader = "x-key"
```
autodidaddict, line 71: I know we're the only ones actually using this interface, but this still feels like a potential collision. Maybe something like `x-keyvalue-key` where `keyvalue` here is the host service to which the header belongs.

I'd also like to see some standardization where all of the header constants either start or end with `header`.

Below, we use `httpUrl` and it looks like it's a variable, whereas `headerHttpUrl` or `httpUrlHeader` might be more self-documenting.

agent/providers/lib/v8.go
```diff
@@ -1250,6 +1216,22 @@ func (v *V8) newObjectStoreObjectTemplate() *v8.ObjectTemplate {
 	return objectStore
 }
 
+func (v *V8) marshalValue(val *v8.Value) ([]byte, error) {
+	if val.IsObject() || val.IsArray() {
+		return val.MarshalJSON()
+	} else if val.IsBigInt() {
```
autodidaddict, line 1222: Will this cause problems when/if endianness changes?

internal/node/services/lib/http.go
```diff
@@ -21,6 +21,8 @@ const httpServiceMethodHead = "head"
 
 const defaultHTTPRequestTimeoutMillis = 2500
 
+const httpURL = "x-http-url"
```
autodidaddict, line 24: is this defined in multiple places?

internal/node/services/lib/keyvalue.go
```diff
@@ -16,6 +16,8 @@ const kvServiceMethodSet = "set"
 const kvServiceMethodDelete = "delete"
 const kvServiceMethodKeys = "keys"
 
+const keyHeader = "x-key"
```
autodidaddict, line 19: this const looks like it's defined in multiple places

autodidaddict (changes requested):

agent/providers/lib/v8.go
kthomas: Agree, will clean this up.

autodidaddict:
Spec suite is failing

agent/providers/lib/v8.go
kthomas: The `IsBigInt` check itself shouldn't cause problems if endianness changes. What we do with the bytes when they are unmarshaled at a later time is something to consider. We are using a string representation with no type hint inside key/value and object stores 🤔

agent/providers/lib/v8.go
```diff
@@ -1029,15 +1001,16 @@ func (v *V8) newMessagingObjectTemplate() *v8.ObjectTemplate {
 			return v.iso.ThrowException(val)
 		}
 
-		var msgresp *agentapi.HostServicesMessagingResponse
-		err = json.Unmarshal(resp.Data, &msgresp)
-		if err == nil && len(msgresp.Errors) > 0 {
-			val, _ := v8.NewValue(v.iso, msgresp.Errors[0])
-			return v.iso.ThrowException(val)
-		}
+		// var msgresp *agentapi.HostServicesMessagingResponse
```
autodidaddict, line 1004: if this commented code is no longer needed, we should remove it

agent/providers/lib/v8.go
```diff
@@ -1250,6 +1215,62 @@ func (v *V8) newObjectStoreObjectTemplate() *v8.ObjectTemplate {
 	return objectStore
 }
 
+func (v *V8) marshalValue(val *v8.Value) ([]byte, error) {
```
autodidaddict, line 1218: this still looks like we're attempting to marshal primitives on the way to the host service. Checking for the JSON object is fine, but we shouldn't be marshaling booleans or numbers or bigints. In the JavaScript code, the developer should only be able to use JSON or strings as values when submitting requests.

Seems like the bigint, boolean, and number branches of this statement should be removed.

Am I misinterpreting the purpose of this code?

autodidaddict (changes requested):

agent/providers/lib/v8.go
kthomas: The developer can only use JSON or strings as values when submitting request based on how this code is written.

autodidaddict: as discussed in another forum, we shouldn't be marshaling at all and should just be dealing with byte arrays, where the developer is responsible for encoding.

internal/agent-api/client.go
```diff
@@ -23,7 +23,16 @@ type EventCallback func(string, cloudevents.Event)
 type LogCallback func(string, LogEntry)
 
 const (
-	nexTriggerSubject = "x-nex-trigger-subject"
+	NexTriggerSubject = "x-nex-trigger-subject"
+	NexRuntimeNs      = "x-nex-runtime-ns"
+
+	HttpURLHeader = "x-http-url"
```
autodidaddict, line 29: We should('ve) probably had nex in all of these headers, just to make it harder for accidental collisions. For example `x-nex-http-url` instead of `x-http-url`

autodidaddict (approved):

PR #196, opened 2024-04-23
adds ability for preflight to create config

jordan-rash:


autodidaddict (approved):
This is going to make the CLI too easy to use.

PR #197, opened 2024-04-24
Add support for starting a userland NATS server

### LINKED ISSUE #168: Allow starting a userland NATS server
### Proposed change

The Nex binary already embeds an internal NATS server for private use. It would be cool if we could also have the Nex server start a "regular" (aka userland) NATS server by passing a configuration file.

### Use case

When developing locally it might be more convenient if the NATS node started its own NATS server. Also, on constrained devices it might be easier to just deploy the `nex` binary rather than both `nex` and `nats`

### Contribution

_No response_

kthomas:
This PR allows users to configure and start a "userland" NATS server when starting a node.

Error handling inside of `node.init()` was not failing fast, resulting in discarded errors. This PR includes a fix to use `errors.Join` which allows the node to gracefully abort initialization using `shutdown()` instead of panicking.

Closes #168.

### DIFF
--- a/internal/node/node.go
+++ b/internal/node/node.go
@@ -55,7 +55,8 @@ type Node struct {
 	keypair   nkeys.KeyPair
 	publicKey string
 
-	nc *nats.Conn
+	natspub *server.Server
+	nc      *nats.Conn
 
 	natsint *server.Server
 	ncint   *nats.Conn
@@ -234,6 +235,15 @@ func (n *Node) init() error {
 			n.log.Info("Telemetry status", slog.Bool("metrics", n.config.OtelMetrics), slog.Bool("traces", n.config.OtelTraces))
 		}
 
+		// start public NATS server
+		err = n.startPublicNATS()
+		if err != nil {
+			n.log.Error("Failed to start public NATS server", slog.Any("err", err))
+			err = fmt.Errorf("failed to start public NATS server: %s", err)
+		} else {
+			n.log.Info("Public NATS server started", slog.String("client_url", n.natspub.ClientURL()))
+		}
+
 		// setup NATS connection
 		n.nc, err = models.GenerateConnectionFromOpts(n.opts)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


autodidaddict:
It would be great if there was another JSON file in the sample configs section that showed the public NATS configuration

kthomas:
Should we validate that the provided options for the public NATS connection match the public NATS server config? As it currently stands, the code will allow a public NATS server to be configured and the NATS connection could still connect successfully to a _different_ NATS server.

internal/node/node.go
```diff
@@ -218,52 +223,68 @@ func (n *Node) generateKeypair() error {
...
 			n.log.Info("Established node NATS connection", slog.String("servers", n.opts.Servers))
 		}
 
-		// init internal NATS server
-		err = n.startInternalNATS()
-		if err != nil {
-			n.log.Error("Failed to start internal NATS server", slog.Any("err", err))
-			err = fmt.Errorf("failed to start internal NATS server: %s", err)
+		// start internal NATS server
+		_err = n.startInternalNATS()
```
autodidaddict, line 264: this looks a little awkward. The underscore prefix typically indicates that the variable isn't being used, but this variable is used in multiple places.

internal/node/node.go
```diff
@@ -218,52 +223,68 @@ func (n *Node) generateKeypair() error {
...
 		}
 
-		go n.manager.Start()
-
-		// init API listener
-		n.api = NewApiListener(n.log, n.manager, n)
-		err = n.api.Start()
-		if err != nil {
-			n.log.Error("Failed to start API listener", slog.Any("err", err))
+		if err == nil {
```
autodidaddict, line 278: could this just be an else clause to the previous `if _err != nil` ?

autodidaddict (approved):
The use of `_err` instead of `err` is a bit jarring to me, but I don't think it's reason enough to stop the PR

PR #199, opened 2024-04-25
Multiple logger options

jordan-rash:

### DIFF
--- a/internal/models/cli.go
+++ b/internal/models/cli.go
@@ -52,6 +52,8 @@ type Options struct {
 	TlsFirst bool
 	// Namespace for scoping workload requests
 	Namespace string
+	// Type of logger
+	Logger string
 	// LogLevel is the log level to use
 	LogLevel string
 	// LogJSON enables JSON logging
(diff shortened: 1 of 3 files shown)


jordan-rash:
It its current form, the nats connection is severed before the last 2-3 logs get published.  @kthomas has an idea for how to fix, we are currently looking into it

autodidaddict:
> It its current form, the nats connection is severed before the last 2-3 logs get published. @kthomas has an idea for how to fix, we are currently looking into it

Calling `Flush` after doing a publish forces a round-trip to the server which then forces that last publication to make it to the server. I just don't know if this has any kind of performance impact (the nats publishes should probably be done from an unbuffered channel so logging doesn't slow the app down)

jordan-rash:
> > It its current form, the nats connection is severed before the last 2-3 logs get published. @kthomas has an idea for how to fix, we are currently looking into it
> 
> Calling `Flush` after doing a publish forces a round-trip to the server which then forces that last publication to make it to the server. I just don't know if this has any kind of performance impact (the nats publishes should probably be done from an unbuffered channel so logging doesn't slow the app down)

I've already tried that...there is something exiting the program before the main goroutine is finished.  I took that code of of the PR, but something about the "graceful shutdown" its as ballarina like as we would hope

kthomas (approved):

PR #202, opened 2024-04-29
Build and run on windows

kthomas:
This PR adds additional build tags to ensure `nex` can build and run on Windows.


internal/node/node_signals_linux.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"os"
+	"os/signal"
+	"syscall"
+)
+
+func (n *Node) installSignalHandlers() {
+	n.log.Debug("installing signal handlers")
+	// both firecracker and the embedded NATS server(s) register signal handlers... wipe those so ours are the ones being used
+	signal.Reset(syscall.SIGINT, syscall.SIGTERM, syscall.SIGUSR1, syscall.SIGUSR2, syscall.SIGHUP)
```
autodidaddict, line 14: the only difference between linux and windows here is that linux wipes `SIGUSR1` and `SIGUSR2` .. do we need this or can this code be shared?

kthomas: No dice :/

No `SIGUSR1` or `SIGUSR2` on Windows.

nex/node_noop.go
jordan-rash, line 1: i would rename this node_darwin.go

autodidaddict:
Rather than have us continue to get pings every time you commit... can you leave this in draft until it's ready for review?

agent/agent.go
```diff
@@ -177,6 +184,10 @@ func (a *Agent) cacheExecutableArtifact(req *agentapi.DeployRequest) (*string, e
 	fileName := fmt.Sprintf("workload-%s", *a.md.VmID)
 	tempFile := path.Join(os.TempDir(), fileName)
 
+	if strings.EqualFold(runtime.GOOS, "windows") && strings.EqualFold(*req.WorkloadType, "elf") {
```
autodidaddict, line 187: I think it would be safer and more predictable if we avoided these kinds of runtime queries and instead used build tags and, better yet, cross-platform libraries, to avoid all of these branches.

internal/node/config.go
```diff
@@ -26,14 +28,16 @@ func LoadNodeConfiguration(configFilepath string) (*models.NodeConfiguration, er
 		config.WorkloadTypes = models.DefaultWorkloadTypes
 	}
 
-	// TODO-- audit for *string
+	if strings.EqualFold(runtime.GOOS, "windows") && !config.NoSandbox {
```
autodidaddict, line 31: see previous thing that makes me worried... runtime check for windows os rather than using build tags and/or xplat library

internal/node/controlapi.go
```diff
@@ -69,40 +92,52 @@ func (api *ApiListener) PublicXKey() string {
...
-	}
-	_, err = api.node.nc.Subscribe(controlapi.APIPrefix+".PING", api.handlePing)
+	var sub *nats.Subscription
+	var err error
+
+	sub, err = api.node.nc.Subscribe(controlapi.APIPrefix+".PING", api.handlePing)
 	if err != nil {
 		api.log.Error("Failed to subscribe to ping subject", slog.Any("err", err), slog.String("id", api.PublicKey()))
 	}
+	api.subz = append(api.subz, sub)
```
autodidaddict, line 102: while there's nothing wrong with us manually maintaining a list of subscriptions so that we can un-sub from them later... that's what the `Drain` command on the connection does. All we have to do is drain that once and all the subscriptions that came from that client are disposed of, and likely far faster than doing it 1-by-1.

internal/node/node.go
```diff
@@ -195,13 +185,17 @@ func (n *Node) createPid() error {
...
-		if err != nil {
+		if err != nil && !strings.EqualFold(runtime.GOOS, "windows") {
+			n.log.Warn("failed to lookup running process by pid", slog.Int("pid", pid), slog.String("err", err.Error()))
 			return err
 		}
 
-		err = process.Signal(syscall.Signal(0))
-		if err == nil {
-			return fmt.Errorf("node process already running; pid: %d", pid)
+		keepExistingPid := strings.EqualFold(os.Getenv("NEX_ENVIRONMENT"), "spec") // HACK!!! there must be a better way 💀
```
autodidaddict, line 193: this looks like an example of adding code into the node in order to accommodate a test peculiarity?

internal/node/processmanager/process_mgr_windows.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+// Initialize an appropriate agent process manager instance based on the sandbox config value
+func NewProcessManager(
+	log *slog.Logger,
+	config *models.NodeConfiguration,
+	telemetry *observability.Telemetry,
+	ctx context.Context,
+) (ProcessManager, error) {
+	if config.NoSandbox {
```
autodidaddict, line 21: this is an if statement that should never return true. Sandbox _can't_ be enabled on windows or mac/darwin.

internal/node/controlapi.go
kthomas: There is an order-of-operations issue with regard to shutting down the machine manager and draining the NATS connections. I think we can remove this manual management of API subscriptions if we change order of operations in `shutdown()`. Will look into this.

agent/agent.go
kthomas: Not a fan of this pattern at all. Will make all of these instances live behind build tags instead, although this may be a little uglier.

internal/node/node.go
kthomas: Yeah, this smells.

agent/agent.go
kthomas: To be addressed in a subsequent, small PR ;)

internal/node/node.go
kthomas: To be addressed in a subsequent, small PR ;)

autodidaddict:
I can't digest a PR of this size. I think the only option is to merge it and then deal with the consequences in multiple smaller PRs after this.

So let's call my approval of this PR a begrudging approval.

autodidaddict (approved):
:sob:

PR #220, opened 2024-05-09
formalizes the host services contract, creates explicit client and server

autodidaddict:
It looks like a lot of changes, but mostly it's just been relocating and simplifying code. I also created a standard envelope used for responses from host services so that it would be easy to tell errors from success without knowing the individual service's payload type.

Because of the new interface used for host services, the `request many` function on messaging is currently commented out. Rather than delaying all this work for the one function, I'm okay with leaving it disabled in main for a little while until I can come up with a good way of supporting this kind of functionality for host services.


autodidaddict:
Unsure why the tests are failing, and it looks like they're not failing as a result of the new code here. Given people's lack of availability this week, I wonder if it makes sense to merge this and then fix the tests in a separate effort?

autodidaddict:
This addresses #170

kthomas (approved):

PR #222, opened 2024-05-14
Add tracing to host services

### LINKED ISSUE #132: Add host services tracing spans
### Proposed change

Right now the open telemetry tracing support is limited to just the execution trigger handler on functions. We want to expand this so that we can create child spans that maintain the appropriate parent for all of the host services, including kv, object store, and messaging.

### Use case

We want to see child spans every time a function utilizes a host service capability

### Contribution

_No response_

kthomas:
This PR adds otel child span support for tracing host services execution.

image

Closes #132.

### DIFF
--- a/agent/providers/api.go
+++ b/agent/providers/api.go
@@ -3,6 +3,7 @@ package providers
 import (
 	"errors"
 
+	"github.com/nats-io/nats.go"
 	"github.com/synadia-io/nex/agent/providers/lib"
 	agentapi "github.com/synadia-io/nex/internal/agent-api"
 )
@@ -27,7 +28,7 @@ type ExecutionProvider interface {
 	Deploy() error
 
 	// Execute a deployed function, if supported by the execution provider implementation (e.g., "v8" and "wasm" types)
-	Execute(subject string, payload []byte) ([]byte, error)
+	Execute(headers nats.Header, payload []byte) ([]byte, error)
 
 	// Undeploy a workload, giving it a chance to gracefully clean up after itself (if applicable)
 	Undeploy() error
--- a/agent/providers/api.go
+++ b/agent/providers/api.go
@@ -38,14 +38,6 @@ type ExecutionProvider interface {
 	Validate() error
 }
 
-// HACK!!! theory was otel was not returning a propagator... but this didn't help...
-// func init() {
-// 	otel.SetTextMapPropagator(propagation.NewCompositeTextMapPropagator(
-// 		propagation.TraceContext{},
-// 		propagation.Baggage{},
-// 	))
-// }
-
 // NewExecutionProvider initializes and returns an execution provider for a given work request
 func NewExecutionProvider(params *agentapi.ExecutionProviderParams) (ExecutionProvider, error) {
 	if params.WorkloadType == nil {
--- a/agent/providers/api.go
+++ b/agent/providers/api.go
@@ -1,9 +1,9 @@
 package providers
 
 import (
+	"context"
 	"errors"
 
-	"github.com/nats-io/nats.go"
 	"github.com/synadia-io/nex/agent/providers/lib"
 	agentapi "github.com/synadia-io/nex/internal/agent-api"
 )
@@ -28,7 +28,7 @@ type ExecutionProvider interface {
 	Deploy() error
 
 	// Execute a deployed function, if supported by the execution provider implementation (e.g., "v8" and "wasm" types)
-	Execute(headers nats.Header, payload []byte) ([]byte, error)
+	Execute(ctx context.Context, payload []byte) ([]byte, error)
 
 	// Undeploy a workload, giving it a chance to gracefully clean up after itself (if applicable)
 	Undeploy() error
@@ -37,6 +37,14 @@ type ExecutionProvider interface {
 	// statically-linked binary or raw source code, depending on provider implementation
 	Validate() error
 }
+
+// HACK!!! theory was otel was not returning a propagator... but this didn't help...
+// func init() {
+// 	otel.SetTextMapPropagator(propagation.NewCompositeTextMapPropagator(
+// 		propagation.TraceContext{},
+// 		propagation.Baggage{},
+// 	))
+// }
 
 // NewExecutionProvider initializes and returns an execution provider for a given work request
 func NewExecutionProvider(params *agentapi.ExecutionProviderParams) (ExecutionProvider, error) {
--- a/agent/providers/lib/v8.go
+++ b/agent/providers/lib/v8.go
@@ -100,7 +100,7 @@ func (v *V8) Deploy() error {
 	subject := fmt.Sprintf("agentint.%s.trigger", v.vmID)
 	_, err := v.nc.Subscribe(subject, func(msg *nats.Msg) {
 		ctx := otel.GetTextMapPropagator().Extract(context.Background(), propagation.HeaderCarrier(msg.Header))
-		ctx = context.WithValue(ctx, agentapi.NexTriggerSubject, subject) //nolint:all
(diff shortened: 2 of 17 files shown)


agent/providers/api.go
```diff
@@ -27,7 +28,7 @@ type ExecutionProvider interface {
 	Deploy() error
 
 	// Execute a deployed function, if supported by the execution provider implementation (e.g., "v8" and "wasm" types)
-	Execute(subject string, payload []byte) ([]byte, error)
+	Execute(headers nats.Header, payload []byte) ([]byte, error)
```
autodidaddict, line 31: let's make this a `context.Context` here, as is the Go convention. We shouldn't be bleeding the nats.Header type into the interface for a host services provider client.

autodidaddict: we can use `context.WithValue(parent_ctx, key, value)` to produce a context that has the information the provider needs, like the trigger subject that originally instigated the call.

https://pkg.go.dev/context#WithValue

agent/providers/lib/v8.go
```diff
@@ -97,7 +97,7 @@ func (v *V8) Deploy() error {
 	subject := fmt.Sprintf("agentint.%s.trigger", v.vmID)
 	_, err := v.nc.Subscribe(subject, func(msg *nats.Msg) {
 		startTime := time.Now()
-		val, err := v.Execute(msg.Header.Get(agentapi.NexTriggerSubject), msg.Data)
+		val, err := v.Execute(msg.Header, msg.Data)
```
autodidaddict, line 100: produce a context here. First, create a context by extracting the otel header from `msg.Header`, which is done via https://pkg.go.dev/go.opentelemetry.io/otel/propagation#TraceContext.Extract 

Calling extract on the nats message header will produce a `context.Context` . None of our code should ever manually copy a traceparent into an explicitly named header variable.

autodidaddict: so:
1. create a context via otel extract from the nats header
1. create a new context from the previous by adding the trigger subject value to it

agent/providers/lib/v8.go
```diff
@@ -108,7 +108,8 @@ func (v *V8) Deploy() error {
 		err = msg.RespondMsg(&nats.Msg{
 			Data: val,
 			Header: nats.Header{
-				agentapi.NexRuntimeNs: []string{strconv.FormatInt(runtimeNanos, 10)},
+				agentapi.NexRuntimeNs:    []string{strconv.FormatInt(runtimeNanos, 10)},
+				agentapi.OtelTraceparent: msg.Header.Values(agentapi.OtelTraceparent),
```
autodidaddict, line 112: never manually add a traceparent field to a nats header. Always use the current context and inject, which takes a trace context and does the header injection for you (keeping that operation opaque) 

https://pkg.go.dev/go.opentelemetry.io/otel/propagation#TraceContext.Inject

agent/providers/lib/v8.go
```diff
@@ -127,12 +128,20 @@ func (v *V8) Deploy() error {
 // Trigger execution of the deployed function; expects a `Validate` to have succeeded and `ubs` to be non-nil.
 // The executed function can optionally return a value, in which case it will be deemed a reply and returned
 // to the caller. In the case of a nil or empty value returned by the function, no reply will be sent.
-func (v *V8) Execute(subject string, payload []byte) ([]byte, error) {
+func (v *V8) Execute(headers nats.Header, payload []byte) ([]byte, error) {
```
autodidaddict, line 131: context not headers

agent/providers/lib/v8.go
```diff
@@ -127,12 +128,20 @@ func (v *V8) Deploy() error {
...
 	if v.ubs == nil {
 		return nil, fmt.Errorf("invalid state for execution; no compiled code available for vm: %s", v.name)
 	}
 
-	ctx, err := v.newV8Context()
+	subject := headers.Get(agentapi.NexTriggerSubject)
+
+	var traceID *string
```
autodidaddict, line 138: use extract to a context here

host-services/builtins/builtins_client.go
```diff
@@ -43,11 +47,15 @@ func (c *BuiltinServicesClient) KVGet(key string) ([]byte, error) {
...
 }
 
-func (c *BuiltinServicesClient) KVSet(key string, value []byte) (*agentapi.HostServicesKeyValueResponse, error) {
+func (c *BuiltinServicesClient) KVSet(key string, value []byte, traceID *string) (*agentapi.HostServicesKeyValueResponse, error) {
 	metadata := map[string]string{
 		agentapi.KeyValueKeyHeader: key,
 	}
 
+	if traceID != nil {
+		metadata[agentapi.OtelTraceparent] = *traceID
```
autodidaddict, line 56: again, this belongs in a context which can/should then be passed to the builtins client

host-services/builtins/builtins_client.go
```diff
@@ -65,11 +73,15 @@ func (c *BuiltinServicesClient) KVSet(key string, value []byte) (*agentapi.HostS
 	return &kvResponse, nil
 }
 
-func (c *BuiltinServicesClient) KVDelete(key string) (*agentapi.HostServicesKeyValueResponse, error) {
+func (c *BuiltinServicesClient) KVDelete(key string, traceID *string) (*agentapi.HostServicesKeyValueResponse, error) {
```
autodidaddict, line 76: remove the trace ID from all of these functions.

autodidaddict (changes requested):
In summary:
1. always use context to go to/from a `msg.Header`
2. never use trace ID as a method parameter

host-services/builtins/builtins_client.go
autodidaddict: I'd imagine this should be `KVDelete(ctx context.Context, key string) (...)` and then when it calls the generic host services client `PerformRpc` that should also take a context, e.g. `PerformRpc(ctx context.Context, ...)` when perform RPC is crafting a messaging, it should then use the trace context Inject method to copy all relevant trace data from the context to the NATS headers.

We should not have a constant for the header key used for tracing, and should always use context round tripping and never manually read/write the traceparent header.

Using the context is idiomatic and it also allows us to add values and information later without changing the signature (this is why like 90% of Go libraries accept a context as the first method param)

autodidaddict:
For the traces that currently are using the `agentint.>` subject as the name, instead let's do the following:

* call it "host services call"
* Add the following as fields to the span: `workload_id`, `service`, `method`

agent/providers/api.go
```diff
@@ -37,6 +38,14 @@ type ExecutionProvider interface {
 	Validate() error
 }
 
+// HACK!!! theory was otel was not returning a propagator... but this didn't help...
```
autodidaddict, line 41: let's remove the commented out stuff now that it works

host-services/server.go
```diff
@@ -81,17 +87,39 @@ func (h *HostServicesServer) handleRPC(msg *nats.Msg) {
...
+	_, span := h.tracer.Start(ctx, "host services call",
+		trace.WithSpanKind(trace.SpanKindClient),
+		trace.WithAttributes(
+			attribute.String("workload_id", vmID),
+			attribute.String("service", serviceName),
+			attribute.String("method", method),
+		))
+	defer span.End()
+
+	span.AddEvent("RPC request")
```
autodidaddict, line 101: typically when we add events we use a past tense form. This way I can tell the difference between whether this event marks the beginning or the end of an RPC request.

autodidaddict: If we have an "RPC Request Began" and "RPC Request Completed" event, and then we flag the span as error for a fail, it becomes a pinch easier to find what we're looking for in a grafana search

jordan-rash (approved):

PR #231, opened 2024-05-20
Select random node as devrun target

kthomas:


nex/devrunner.go
```diff
@@ -136,6 +134,33 @@ func RunDevWorkload(ctx context.Context, logger *slog.Logger) error {
 	return nil
 }
 
+func randomNode(nodeClient *controlapi.Client) (*controlapi.PingResponse, error) {
+	candidates, err := listNodes(nodeClient)
+	if err != nil {
+		return nil, err
+	}
+
+	shuffled := make([]*controlapi.PingResponse, len(candidates))
```
autodidaddict, line 143: rather than shuffling the list, why not just generate a random number between 0 and the number of candidates-1 and use that as the index?

autodidaddict (changes requested):

autodidaddict (approved):

PR #235, opened 2024-05-21
adding validation around binary types

### LINKED ISSUE #233: Attempt to Validate Binaries
### Proposed change

We are not currently validating binaries are being deployed to machines that support the binaries architecture.  For example, if we attempt to deploy a linux/arm64 binary to a no-sandbox windows node, it will attempt to run it.

This issue proposes we attempt to validate a binary as best we can prior to deployment using the file headers.

### Use case

`nex run` and `nex devrun` should error if a binary is attempted to run on a machine with a different OS/Arch

### Contribution

_No response_

jordan-rash:
Closes #233

### DIFF
--- a/nex/devrunner.go
+++ b/nex/devrunner.go
@@ -49,6 +49,14 @@ func init() {
 // Attempts to "deploy a file" by finding a suitable target and publishing the workload to an ad-hoc created bucket
 // and using default issuer and publisher keys stored in ~/.nex. This should be as easy as typing "nex devrun ./amazingapp env1=foo env2=bar"
 func RunDevWorkload(ctx context.Context, logger *slog.Logger) error {
+	if RunOpts.WorkloadType == "elf" || RunOpts.WorkloadType == "native" || true {
+		os, arch, err := validateBinary(DevRunOpts.Filename)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


nex/devrunner.go
```diff
@@ -49,6 +49,14 @@ func init() {
 // Attempts to "deploy a file" by finding a suitable target and publishing the workload to an ad-hoc created bucket
 // and using default issuer and publisher keys stored in ~/.nex. This should be as easy as typing "nex devrun ./amazingapp env1=foo env2=bar"
 func RunDevWorkload(ctx context.Context, logger *slog.Logger) error {
+	if RunOpts.WorkloadType == "elf" || RunOpts.WorkloadType == "native" || true {
```
autodidaddict, line 52: Why is the "or true" on here? Doesn't that make this code block always run?

jordan-rash: it does...because WorkloadType is being determined after the file is uploaded, which isnt going to work here.  ideally, we will required workload for devrun (and default it to elf/native)....i can add that logic

autodidaddict (approved):

kthomas (approved):
🚀

Issue #239, opened 2024-05-22
Smarter Run Command | Node Queries


jordan-rash:
### Proposed change

Currently, when you try to deploy a workload, the nex binary does not validate that the node your sending to can support the particular workload you are providing.  This was originally an intentional design, as `no-sandbox` mode didn't exist and the target OS/Arch was much smaller.  Now, we have to ensure ELFs aren't sent to Windows, arm64 isnt send to arm64, etc.

This issue proposed asking a node if it is able to run a workload prior to tasking.  The "PING" endpoint currently has enough data to assist in this decision, but that puts the task of deciding on the CLI (CLI -> PING -> Parse Responses -> Deploy)

A better approach might be to send a request with details of the workload and have the nodes ACK/NACK if they can support.  This will enable us to go beyond native binaries and as nodes if they can support cetain workloads (not everything can support V8 for example)...tags, etc.

### Use case

Adds more protection to deploying workloads

Depends on: #234

jordan-rash:
Work in #235 to be leveraged with this

autodidaddict:
To add some more specific requirements to this, what we're describing here is the "auction" pattern (this pattern also goes by many other names). One thing with this pattern is we don't want to reuse existing `PING` endpoints because ping's contract implies that any listening node will/must reply. Auction candidates only reply if they meet the criteria.

I propose that we have another endpoint, `$NEX.AUCTION.{namespace}`. Namespace here implies the _intended_ namespace of the candidate workload. 

## Auction Request
Clients making an auction request should supply the following JSON:

```json
{
    "workload_name": "xxx",
    "workload_bytes": "xxx",
    "workload_type": "[v8|wasm|elf/native]",
    "required_tags": [],
}
```

All fields in this payload are required. If there are no required tags, pass an empty list `[]`. Only nodes that think they can run the expected workload candidate should reply. The reply doesn't matter, but it will be the string `OK` when in multi-node auction mode.

Note that we don't need a boolean flag here to indicate acceptance because the existence of a reply on its own constitutes an acceptance. There is _no_ reply to indicate a rejection by the node.

Reasons for rejection (non-response):
* Node has reached workload capacity (number)
* Node has reached workload capacity (bytes/resource consumption)
* Node type mismatch
* Node tags mismatch
* Workload with desired name is already running in the desired namespace on the target node

[...]

PR #242, opened 2024-05-23
Support optional ping request for filtering nodes

kthomas:
Closes https://github.com/synadia-io/nex/issues/239


jordan-rash (approved):
Awesome!

PR #243, opened 2024-05-24
Preliminary support for account security in nats internal server

autodidaddict:
This set of changes will make the internal NATS server multi-tenant safe by provisioning a user/account for each workload, and keeping the workload binary cache in a separate object store (one in each user account).

### DIFF
--- a/internal/node/internal-nats/config_template.go
+++ b/internal/node/internal-nats/config_template.go
@@ -7,12 +7,13 @@ import (
 )
 
 type internalServerData struct {
-	Users             []userData
+	Credentials       map[string]*credentials
 	NexHostUserPublic string
 	NexHostUserSeed   string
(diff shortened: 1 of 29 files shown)


autodidaddict:
Currently in draft. The new internal server struct exists and has tests to verify that it works. Forthcoming commits will have this integrated with the rest of the code.

internal/node/internal-nats/config_template.go
```diff
@@ -33,6 +33,7 @@ accounts: {
 	nexhost: {
 		jetstream: true		
 		users: [
+			{user: nex, password: pass}
```
autodidaddict, line 36: This breaks the design of the internal security.

kthomas (approved):

PR #265, opened 2024-06-08
Per-workload host services connections

autodidaddict:
Each workload is now given its own unique/private host services connection. These connections are created using the following order of configuration (1 being the highest priority)

1. host services configuration on the deployment request
2. host services configuration supplied for the entire node
3. configuration copied from the node's control connection

### DIFF
--- a/host-services/server.go
+++ b/host-services/server.go
@@ -36,6 +36,7 @@ func NewHostServicesServer(ncInternal *nats.Conn, log *slog.Logger, tracer trace
 }
 
 func (h *HostServicesServer) SetHostServicesConnection(workloadId string, nc *nats.Conn) {
+	h.RemoveHostServicesConnection(workloadId)
 	h.hsClientConnections[workloadId] = nc
 }
 
--- a/host-services/server.go
+++ b/host-services/server.go
@@ -15,18 +15,34 @@ import (
 )
 
 type HostServicesServer struct {
-	log      *slog.Logger
-	nc       *nats.Conn
-	services map[string]HostService
-	tracer   trace.Tracer
+	log        *slog.Logger
+	ncInternal *nats.Conn
+	services   map[string]HostService
+	// Every single workload gets its own private host services connection,
+	// even if it's reusing defaults for config
+	hsClientConnections map[string]*nats.Conn
+
+	tracer trace.Tracer
 }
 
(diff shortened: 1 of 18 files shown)


autodidaddict:
It looks like the tests are failing in and around the use of host services (no surprise there). However, when doing manual tests it all works, so I'm wondering if there's just something that needs to be updated in the proxy or fixtures?

autodidaddict:
```
./nex devrun ../examples/v8/echofunction/src/echofunction.js --trigger_subject=foo.trigger --type=v8
Reusing existing issuer account key: /home/kevin/.nex/issuer.nk
Reusing existing publisher xkey: /home/kevin/.nex/publisher.xk
🚀 Workload 'echofunctionjs' accepted. You can now refer to this workload with ID: cpicmsn52omib90596g0 on node NC27GEVVHI6VTT6MOAH64VGZ5VMSYUKB5CWAELUY2SEC3QR4W5ZPJ4JE
kevin in wopr in nex/nex on  feat/perworkloadhsconfig via 🐹 v1.22.2 
❯ nats req 'foo.trigger' 'howdy'
17:23:49 Sending request on "foo.trigger"
17:23:49 Received with rtt 3.293876ms
{"triggered_on":"foo.trigger","payload":"howdy"}
```

host-services/server.go
```diff
@@ -15,18 +15,34 @@ import (
...
-	log      *slog.Logger
-	nc       *nats.Conn
-	services map[string]HostService
-	tracer   trace.Tracer
+	log        *slog.Logger
+	ncInternal *nats.Conn
+	services   map[string]HostService
+	// Every single workload gets its own private host services connection,
+	// even if it's reusing defaults for config
+	hsClientConnections map[string]*nats.Conn
```
kthomas, line 23: My preference is to put the host services connection in the `AgentClient`, alongside the agent's internal NATS connection. Once we get the tests working we can look at how to make that approach place nicely with the host services server and internal nats structs.

autodidaddict: Conceptually the host services client doesn't belong inside the agent client, it's the connection through which HostServices makes requests of the built-ins.

Let's get the tests working and we can move forward from there.

kthomas: This makes sense- the idea was born from my desire to move the internal per-workload internal NATS connection into the `AgentClient` and cache it there instead of caching the credentials inside of the NATS internal server struct.

kthomas (approved):

PR #266, opened 2024-06-09
Ensure subscriptions drain

kthomas:
This PR ensures all subscriptions are drained before exiting.

Also, `node.Stop()` was a method added solely for use by the spec suite. It has been removed in favor of using signals. This should dramatically increase stability of the spec suite.


autodidaddict:
I'm not sure I fully understand what the benefit is from manually draining subscriptions one by one? When the NATS connection is closed, its subscriptions go away. If the server times out a client connection, that connection's subscriptions go away. 

Is there a scenario or a use case (or than a test) that explicit subscription draining addresses?

internal/agent-api/client.go
```diff
@@ -171,6 +171,10 @@ func (a *AgentClient) Drain() error {
 			// no-op for now, try the next one... perhaps we should return the error here in the future?
 		}
 
+		for sub.IsDraining() {
+			time.Sleep(time.Millisecond * 10)
```
autodidaddict, line 175: where did this value come from? Is it always 10 millis or does it change depending on the machine?

kthomas: Currently it's just hard-coded to 10 millis. Since draining is async this blocks until we know the sub has drained. An optimization would be to use a waitgroup to drain concurrently but seemed like a premature optimization.

autodidaddict (approved):

PR #270, opened 2024-06-10
Add DNS support for external name resolution

kthomas:
This PR equips the nex node with an embedded nameserver, running on an ephemeral port, which is provided to sandboxed agent VMs via Firecracker MMDS. The agent uses iptables to setup NAT OUTPUT chain rules for sending local DNS packets destined for port 53 (UDP or TCP) to the configured nameserver.


autodidaddict:
So we weren't able to do this with just configuration? :sob:

autodidaddict:
Is it possible to check the node configuration and only activate the DNS stuff is sandbox is on? If the node isn't using firecracker then running the embedded DNS is unnecessary

kthomas:
> Is it possible to check the node configuration and only activate the DNS stuff is sandbox is on? If the node isn't using firecracker then running the embedded DNS is unnecessary

Yes, and definitely something I wanted to bring up. Will only activate this in sandbox mode.

jordan-rash (approved):
Dope!

PR #274, opened 2024-06-12
updated rootfs with dns softlink

jordan-rash:


kthomas (changes requested):
I would strongly prefer having out of the box support for whatever nameservers are in my `/etc/resolv.conf`. I shouldn't have to change my host nameservers to use nex.

autodidaddict (approved):

autodidaddict (approved):

PR #285, opened 2024-06-18
Adds the ability to deny certain trigger subjects for functions

autodidaddict:
1. Always rejects a certain set of subject spaces for functions: `$SYS.>`, `$JS.>`, and `$NEX.>`. 
2. You can add more trigger subject restrictions in the node config using the `deny_trigger_subjects` field

This is what it looks like to try and devrun a function with a denied trigger subject:

```
❯ ./nex devrun ../examples/v8/echofunction/src/echofunction.js --trigger_subject='$NEX.events.>' --type=v8
Reusing existing issuer account key: /home/kevin/.nex/issuer.nk
Reusing existing publisher xkey: /home/kevin/.nex/publisher.xk
[ERR] 2024-06-18 14:21:56 - failed to devrun workload err=The trigger subject $NEX.events.> overlaps with subject(s) in this node's deny list
```

### DIFF
--- a/internal/node/node.go
+++ b/internal/node/node.go
@@ -230,7 +230,10 @@ func (n *Node) init() error {
 			n.log.Error("Failed to load node configuration file", slog.Any("err", _err), slog.String("config_path", n.nodeOpts.ConfigFilepath))
 			err = errors.Join(err, _err)
 		} else {
-			n.log.Info("Loaded node configuration", slog.String("config_path", n.nodeOpts.ConfigFilepath))
+			n.log.Info("Loaded node configuration",
+				slog.String("config_path", n.nodeOpts.ConfigFilepath),
+				slog.Any("deny_trigger_subjects", n.config.DenyTriggerSubjects),
+			)
 		}
 
 		n.telemetry, _err = observability.NewTelemetry(n.ctx, n.log, n.config, n.publicKey)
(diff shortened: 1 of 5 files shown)


internal/node/node.go
```diff
@@ -230,7 +230,10 @@ func (n *Node) init() error {
 			n.log.Error("Failed to load node configuration file", slog.Any("err", _err), slog.String("config_path", n.nodeOpts.ConfigFilepath))
 			err = errors.Join(err, _err)
 		} else {
-			n.log.Info("Loaded node configuration", slog.String("config_path", n.nodeOpts.ConfigFilepath))
+			n.log.Info("Loaded node configuration",
+				slog.String("config_path", n.nodeOpts.ConfigFilepath),
+				slog.Any("deny_trigger_subjects", n.config.DenyTriggerSubjects),
```
kthomas, line 235: This `deny_trigger_subjects` seemed a little out of place here.

kthomas (approved):
Simple enough-- one comment about the `deny_trigger_subjects` being logged in what feels like an unexpected place. Otherwise LGTM.

internal/node/node.go
autodidaddict: Hehe it's leftover debug. I didn't see it when cleaning because it wasn't a printf ;)

kthomas (approved):

PR #290, opened 2024-06-24
Ensures trigger and host services messaging connections are the same

autodidaddict:
When a trigger connection has been supplied for a given workload, this gives the host services provider implementations the ability to access that "secondary" connection. Right now, the only service that uses this is `messaging`, which ensures that the subject space in which a function is triggered is also the same in which it can publish/request.


autodidaddict:
:thinking: unsure where the `exec format error` message is coming from in the tests:

```
[WARN] 12:48:23 - Agent command failed to start error=fork/exec /Users/runner/go/bin/nex-agent: exec format error
Error:  12:48:23 - Failed to spawn nex-agent for pool error=fork/exec /Users/runner/go/bin/nex-agent: exec format error
```

host-services/builtins/messaging.go
```diff
@@ -55,21 +55,27 @@ func (m *MessagingService) Initialize(config json.RawMessage) error {
...
+	conns []*nats.Conn,
 	namespace string,
 	workloadId string,
 	method string,
 	workloadName string,
 	metadata map[string]string,
 	request []byte) (hostservices.ServiceResult, error) {
 
+	conn := conns[0]
+	// if there's a separate "trigger" connection, we use that to ensure consistency
```
kthomas, line 67: This is a little confusing, that is, that the trigger connection is always at `[1]`.

autodidaddict: yeah I was going to add numeric constants for those indexes but it seemed like overkill for just one.
I had a previous incarnation of this that used multiple discrete connection parameters but I figured that would get ugly if we needed yet another connection.

kthomas (approved):

PR #291, opened 2024-06-24
adds retry/backoff to preflight

jordan-rash:

### DIFF
--- a/internal/node/preflight/init.go
+++ b/internal/node/preflight/init.go
@@ -3,9 +3,10 @@ package preflight
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"io"
+	"log/slog"
 	"net/http"
+	"time"
 
 	"github.com/fatih/color"
 )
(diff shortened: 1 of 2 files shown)


internal/node/preflight/init.go
```diff
@@ -16,39 +17,51 @@ var (
...
 			if err != nil {
-				fmt.Printf("error making http request: %s\n", err)
+				logger.Error("error making http request", slog.Any("err", err))
 				return ""
 			}
 			defer res.Body.Close()
+			if res.StatusCode != 200 && backoff < 5 {
+				slog.Warn("failed to get latest version, incrementing backoff", slog.String("status_code", res.Status), slog.Int("backoff", backoff))
+				backoff++
+				time.Sleep(time.Duration(backoff) * time.Second)
```
autodidaddict, line 42: is the maximum backoff failure time here (15 secs) going to potentially mess up CI testing?

autodidaddict (approved):

internal/node/preflight/init.go
jordan-rash: we are hitting a timeout....so its necessary

autodidaddict: cool. so the backoff is to appease the CI deities rather than anger them. :+1:

jordan-rash: Or anyone using GitHub as an unauthenticated user.  From what I saw in the docs, it's rate limited on IP at 60 requests and hour....if you run our test suite, you use about 25 of those requests every time 🙃

spec/node_nofc_test.go
```diff
@@ -91,6 +92,13 @@ var _ = Describe("nex node", func() {
 				_ = os.Setenv("PATH", fmt.Sprintf("%s:%s", os.Getenv("PATH"), agentPath))
 			}
 		})
+
+		preflightForceOnce.Do(func() {
```
jordan-rash, line 96: This is not a good solution, i will be reverting it

PR #298, opened 2024-06-27
fix path lookup

jordan-rash:

### DIFF
--- a/internal/node/controlapi.go
+++ b/internal/node/controlapi.go
@@ -1,6 +1,7 @@
 package nexnode
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"log/slog"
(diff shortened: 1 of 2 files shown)


autodidaddict (approved):

spec/node_linux_test.go
```diff
@@ -76,7 +76,9 @@ var _ = Describe("nex node", func() {
 		opts = &models.Options{
 			Servers: _fixtures.natsServer.ClientURL(),
 		}
-		nodeOpts = &models.NodeOptions{}
+		nodeOpts = &models.NodeOptions{
+			PreflightInstallVersion: "0.2.5",
```
jordan-rash, line 80: this should keep us from making the rated limited call

internal/node/preflight/validate.go
```diff
@@ -12,58 +12,81 @@ import (
...
+	}
+
+	for _, p := range path {
+		e := filepath.Join(p, binary)
+		f, err := os.Stat(e)
+		if err != nil {
+			continue
+		}
+
+		if f.Mode()&0100 == 0 {
```
jordan-rash, line 57: this might not be windows safe

kthomas (approved):

PR #305, opened 2024-07-01
Add support for executing agent-local workloads

kthomas:
This PR allows the workload deploy `location` to be specified using a `file://` or `nats://` URL scheme.

File-based deploy locations are treated as "agent-local" workload deploy requests, meaning the agent will attempt to resolve the named file as an executable installed on its rootfs. File-based workload deploy requests can only be executed in sandbox mode.

### DIFF
--- a/control-api/run.go
+++ b/control-api/run.go
@@ -7,11 +7,15 @@ import (
 	"fmt"
 	"net/url"
 	"regexp"
+	"strings"
 	"time"
 
 	"github.com/nats-io/jwt/v2"
 	"github.com/nats-io/nkeys"
 )
+
+const WorkloadLocationSchemeFile = "file"
+const WorkloadLocationSchemeNATS = "nats"
 
 type DeployRequest struct {
 	Argv         []string    `json:"argv,omitempty"`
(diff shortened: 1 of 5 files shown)


autodidaddict:
> File-based workload deploy requests can only be executed in sandbox mode.
So there's no way for someone to specify a path to a file that didn't ship with the rootfs or when agents are running without a rootfs?

I can't find the error message in the PR that rejects this.. maybe git's hiding the diff.

control-api/run.go
```diff
@@ -111,6 +115,14 @@ func (request *DeployRequest) Validate() (*jwt.GenericClaims, error) {
 		return nil, fmt.Errorf("workload name claim ('%s') does not match requirements (%s)", claims.Subject, workloadRegex)
 	}
 
+	if request.Essential != nil && *request.Essential && request.WorkloadType != NexWorkloadNative {
+		return nil, errors.New("essential workloads must be native")
+	}
+
+	if !strings.EqualFold(request.Location.Scheme, WorkloadLocationSchemeFile) && !strings.EqualFold(request.Location.Scheme, WorkloadLocationSchemeNATS) {
```
autodidaddict, line 122: I can see we're validating that the scheme is correct.. but where do we validate that a `file://` workload is only supported by firecracker sandboxes?

kthomas: https://github.com/synadia-io/nex/pull/305/files#diff-019cc1c548465bdfa17427d6a8f9a2039c705a13c83a04712330d22658e0ad1aR161-R163

👍🏻

kthomas: In addition to the workload manager checking to ensure we are in a sandbox, the agent also prevents execution: https://github.com/synadia-io/nex/pull/305/files#diff-62b6ad581fe3a3059ae8c85ef0f31dde4092bfdecfa7d6857c470bcacaa8cc8bR185-R187

autodidaddict:
As discussed I think if we put this behind a feature flag like `agent_local_workloads` and default that to `false` and it can only be enabled via node configuration, then this should be good.

Short version is nodes should default to not allowing this behavior and it should be explicitly opt-in to enable

kthomas:
Closing in favor of #325

kthomas:
Closing in favor of #325

PR #307, opened 2024-07-02
Fix stability of workload deployment and process accounting

kthomas:
This PR addresses the following instability issues:

- the internal NATS server had no mutex around the modification of the map that stores creds; in addition, we are now caching the first NATS connection created for the node to communicate with running agents

- `MachineManager` and `ProcessManager` implementations were not synchronized; `MachineManager` now returns data about agents and does not lookup agents/workloads using the `ProcessManager` interface

- deploying workloads using the `controlapi` package now implements logic to retry in the event no agent is available

- the spec harness was ill-equipped to handle testing multiple machines in the pool

- `nex node info` output has been updated to include more data

### DIFF
--- a/internal/node/internal-nats/internal_nats_server.go
+++ b/internal/node/internal-nats/internal_nats_server.go
@@ -3,6 +3,7 @@ package internalnats
 import (
 	"context"
 	"errors"
+	"fmt"
 	"log/slog"
 	"net/url"
 	"os"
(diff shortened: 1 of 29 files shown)


internal/node/internal-nats/internal_nats_server.go
```diff
@@ -213,7 +219,7 @@ func (s *InternalNatsServer) StoreFileForID(id string, bytes []byte) error {
 		return err
 	}
 
-	_, err = bucket.PutBytes(ctx, workloadCacheFileKey, bytes)
+	_, err = bucket.PutBytes(ctx, id, bytes)
```
autodidaddict, line 222: Note that this didn't need to be fixed. The single cache key worked fine because the cache bucket is inside an isolated account. Each agent has its own nkey which connects to its own isolated account.

kthomas:
Closing in favor of #324

PR #313, opened 2024-07-08
adds support for pluggable workload types

autodidaddict:
Documentation will have to be updated for this (taken care of in a separate PR). The short version is this:

* You can now supply any workload type name in the node configs `workload_types`, _so long as_ that type is on disk (relative to the agent, e.g. rootfs) in `{agent_plugin_path}/{workload_type}.[so|dylib]`. This means that if you created a new JavaScript provider based on an alternate engine, you might see it in `{agent_plugin_path}/superv8.so`.
* If you want to enable dynamically loaded workload execution providers, supply the `agent_plugin_path` field in the node configuration. If this field exists, plugins are enabled and can be found _only_ at that path. The agent will not look elsewhere for plugins
* If you enable plugins, you will need to build a dynamically linked version of the agent (our `Taskfile` builds a static one), as Go doesn't allow statically linked binaries to dynamically load libraries at runtime.


jordan-rash:
this should be good except for the macOS tests....they are failing for some reason that I didnt have time to investigate

kthomas:
Looks pretty good... can lend a hand on the tests as necessary.

autodidaddict:
> this should be good except for the macOS tests....they are failing for some reason that I didnt have time to investigate

This is what I found when I went spelunking:

```
  [FAILED] Expected
      <*errors.errorString | 0x14000f2ef40>: 
      Failed to deploy workload: workload rejected by agent: Failed to get and write workload artifact to temp dir: nats: jetstream not enabled
      {
          s: "Failed to deploy workload: workload rejected by agent: Failed to get and write workload artifact to temp dir: nats: jetstream not enabled",
      }
  to be nil
  In [It] at: /Users/runner/work/nex/nex/spec/node_nofc_test.go:458 @ 07/08/24 22:54:05.296
  ```
Of course I have no idea what that has to do with any of the changes I made to the code.

kthomas (approved):

PR #316, opened 2024-07-10
adds the ability to add pluggable host services

autodidaddict:
This adds the ability to add a non-builtin host service through node configuration by indicating the plugin path for it. The only way for a function to consume non-builtin services is by making the `hostServices.raw.request` call. If someone wants to create their own host services provider, they could write their own wrapper around the raw rpc.


autodidaddict:
there have been enough changes to how host services work that I'm going to close this in favor of a slightly different approach for enabling this going forward

PR #317, opened 2024-07-10
Adds Dockerfile 

jordan-rash:
This will only work in no-sandbox mode

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ WORKDIR /agent
 COPY go.mod ./
 RUN go mod download
 COPY . .
-RUN go build -tags netgo -ldflags '-extldflags "-static"' -o nexagent ./agent/cmd/nex-agent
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+RUN go mod download
+COPY . .
+RUN go build -o nexcli ./nex
+
+FROM golang AS agentbuilder
+WORKDIR /agent
+COPY go.mod ./
+RUN go mod download
+COPY . .
+RUN go build -tags netgo -ldflags '-extldflags "-static"' -o nexagent ./agent/cmd/nex-agent
```
autodidaddict, line 13: if we're in no sandbox mode, then we don't need to build the agent statically (static build also prevents workload type plugins)

jordan-rash: I can remove that....it was a copy pasta job

autodidaddict (dismissed):

autodidaddict (approved):

PR #324, opened 2024-07-12
Fix stability of workload deployment and process accounting

kthomas:
This PR addresses the following instability issues:

- the internal NATS server had no mutex around the modification of the map that stores creds; in addition, we are now caching the first NATS connection created for the node to communicate with running agents

- MachineManager and ProcessManager implementations were not synchronized; MachineManager now returns data about agents and does not lookup agents/workloads using the ProcessManager interface

- deploying workloads using the controlapi package now implements logic to retry in the event no agent is available

- the spec harness was ill-equipped to handle testing multiple machines in the pool

- nex node info output has been updated to include more data

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -26,11 +26,11 @@ import (
 )
 
 const (
-	defaultAgentHandshakeTimeoutMillis  = 1000
+	defaultAgentHandshakeAttempts       = 5
+	defaultAgentHandshakeTimeoutMillis  = 500
 	runloopSleepInterval                = 250 * time.Millisecond
 	runloopTickInterval                 = 2500 * time.Millisecond
-	workloadExecutionSleepTimeoutMillis = 1000
-	workloadCacheFileKey                = "workload"
+	workloadExecutionSleepTimeoutMillis = 100
 )
 
 // Agent facilitates communication between the nex agent running in the firecracker VM
@@ -46,6 +46,7 @@ type Agent struct {
 	sigs    chan os.Signal
 
 	provider providers.ExecutionProvider
+	subz     []*nats.Subscription
 
 	cacheBucket nats.ObjectStore
 	md          *agentapi.MachineMetadata
@@ -84,6 +85,7 @@ func NewAgent(ctx context.Context, cancelF context.CancelFunc) (*Agent, error) {
 		sandboxed: isSandboxed(),
 		md:        metadata,
 		started:   time.Now().UTC(),
+		subz:      make([]*nats.Subscription, 0),
 	}, nil
 }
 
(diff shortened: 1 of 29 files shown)


agent/agent.go
```diff
@@ -197,6 +197,33 @@ func (a *Agent) cacheExecutableArtifact(req *agentapi.DeployRequest) (*string, e
 	return &tempFile, nil
 }
 
+// deleteExecutableArtifact deletes the installed workload executable
+// and purges it from the internal object store
+func (a *Agent) deleteExecutableArtifact() error {
+	fileName := fmt.Sprintf("workload-%s", *a.md.VmID)
+	tempFile := path.Join(os.TempDir(), fileName)
+
+	// if strings.EqualFold(runtime.GOOS, "windows") && req.WorkloadType == controlapi.NexWorkloadNative {
```
autodidaddict, line 206: if the code commented out here is no longer in use, please delete

agent/agent.go
```diff
@@ -84,6 +85,7 @@ func NewAgent(ctx context.Context, cancelF context.CancelFunc) (*Agent, error) {
 		sandboxed: isSandboxed(),
 		md:        metadata,
 		started:   time.Now().UTC(),
+		subz:      make([]*nats.Subscription, 0),
```
autodidaddict, line 88: I still maintain that manually and explicitly unsubscribing every subscription isn't/shouldn't be necessary, but as long as it doesn't actually make things worse, I'm okay with it.

internal/agent-api/client.go
```diff
@@ -241,6 +248,24 @@ func (a *AgentClient) UptimeMillis() time.Duration {
...
 
+func (a *AgentClient) DeployRequest() *DeployRequest {
+	return a.deployRequest
+}
+
+func (a *AgentClient) IsSelected() bool {
+	return a.selected
+}
+
+func (a *AgentClient) MarkSelected() {
```
autodidaddict, line 259: if these properties shouldn't be set anywhere outside that function, then we should move these property sets into that function and not give developers the chance to footgun

internal/node/processmanager/firecracker_procman.go
```diff
@@ -225,14 +239,14 @@ func (f *FirecrackerProcessManager) StopProcess(workloadID string) error {
 	return nil
 }
 
-func (f *FirecrackerProcessManager) Lookup(workloadID string) (*agentapi.DeployRequest, error) {
-	if request, ok := f.deployRequests[workloadID]; ok {
-		return request, nil
-	}
+// func (f *FirecrackerProcessManager) Lookup(workloadID string) (*agentapi.DeployRequest, error) {
```
autodidaddict, line 242: remove

internal/node/processmanager/spawn_procman.go
```diff
@@ -197,22 +206,22 @@ func (s *SpawningProcessManager) StopProcess(workloadID string) error {
...
 	return nil
 }
 
 // Looks up an agent process. A non-existent agent process returns (nil, nil), not
 // an error
-func (s *SpawningProcessManager) Lookup(workloadID string) (*agentapi.DeployRequest, error) {
-	if request, ok := s.deployRequests[workloadID]; ok {
-		return request, nil
-	}
+// func (s *SpawningProcessManager) Lookup(workloadID string) (*agentapi.DeployRequest, error) {
```
autodidaddict, line 217: remove commented code

nex/nodes.go
```diff
@@ -96,6 +96,7 @@ func renderNodeInfo(info *controlapi.InfoResponse, id string, full bool) {
 	cols.AddRowf("Xkey", info.PublicXKey)
 	cols.AddRow("Version", info.Version)
 	cols.AddRow("Uptime", info.Uptime)
+	cols.AddRow("Available Agents", info.AvailableAgents)
```
autodidaddict, line 99: in the future, this is the kind of thing that's an example of a change that should be in another PR. if this PR is about stability fixes, it should contain stability fixes. This is a control interface API change.

autodidaddict (changes requested):
just a couple cleanup suggestions

internal/agent-api/client.go
kthomas: They are used in another package.

autodidaddict: Having to comment something by saying that it should never be called except by one other function is a potential smell. Months from now when we don't know why we did any of this, we could get tripped up here.

kthomas: Comment removed.

autodidaddict (approved):

PR #325, opened 2024-07-12
Add support for executing agent-local workloads

kthomas:
This PR allows the workload deploy location to be specified using a file:// or nats:// URL scheme.

File-based deploy locations are treated as "agent-local" workload deploy requests, meaning the agent will attempt to resolve the named file as an executable installed on its rootfs. File-based workload deploy requests can only be executed in sandbox mode.


autodidaddict:
This is still something that needs to be behind a feature/configuration flag that defaults to off/unsupported.

PR #328, opened 2024-07-15
Stability Testing

jordan-rash:


nex/runner.go
```diff
@@ -84,26 +89,47 @@ func RunWorkload(ctx context.Context, logger *slog.Logger) error {
...
+
 	argv := []string{}
 	if len(RunOpts.Argv) > 0 {
 		argv = strings.Split(RunOpts.Argv, " ")
 	}
 
 	request, err := controlapi.NewDeployRequest(
 		controlapi.Argv(argv),
-		controlapi.Location(RunOpts.WorkloadUrl.String()),
+		controlapi.Hash(info.Digest),
```
autodidaddict, line 120: Note that this hash format is not the sanitized one that we like... it contains the `SHA-256:` prefix and I think it might even be in a different base. Might be worth running this through the hash "cleaner" that I think we used somewhere else (possibly in the CLI?)

test/stop_test.go
```diff
@@ -21,7 +21,7 @@ func TestStopValidation(t *testing.T) {
 		WorkloadName("testworkload"),
 		WorkloadType("elf"),
 		WorkloadDescription("testy mctesto"),
-		Checksum("hashbrowns"),
+		Hash("hashbrowns"),
```
autodidaddict, line 24: akshually, this should be `Hash("browns")` :)

control-api/events.go
```diff
@@ -50,12 +50,12 @@ type NodeStoppedEvent struct {
...
 
-// TODO: remove omitempty in next version bump
 type HeartbeatEvent struct {
-	Version         string            `json:"version"`
-	NodeID          string            `json:"node_id"`
-	Nexus           string            `json:"nexus,omitempty"`
-	Uptime          string            `json:"uptime"`
-	Tags            map[string]string `json:"tags,omitempty"`
-	RunningMachines int               `json:"running_machines"`
+	AllowDuplicateWorkloads *bool             `json:"allow_duplicate_workloads"`
```
autodidaddict, line 54: Not a fan of using a pointer to a boolean here. Bool's zero value is false, which works fine

internal/models/config.go
```diff
@@ -121,6 +122,10 @@ type AutostartDeployRequest struct {
 func (c *NodeConfiguration) Validate() bool {
 	c.Errors = make([]error, 0)
 
+	if c.AllowDuplicateWorkloads == nil {
```
autodidaddict, line 125: we can get rid of this check and not use a pointer. The allow duplicate workloads field shouldn't be required as it has reasonable defaults (False)

autodidaddict (changes requested):
In this situation (and I can't believe I'm saying this), the pointer to a bool makes sense... however, I noticed we're not using `omit empty` in all the cases where we have the new *bool field. we should do that so that old configurations aren't going to get rejected.

kthomas (dismissed):

kthomas (approved):

PR #333, opened 2024-07-31
Native Workload + Host Services Example

### LINKED ISSUE #218: Create host services example for service-type workload
### Proposed change

Create an example that shows how a full service workload (`elf`) can make use of the same host services that a JavaScript function might

### Use case

People might want their long-running services to have access to the same convenient pre-provisioned resources that functions have

### Contribution

_No response_

jordan-rash:
Closes #218

### DIFF
--- a/examples/native_hostservice/main.go
+++ b/examples/native_hostservice/main.go
@@ -1,0 +1,26 @@
+package main
+
+import (
+	"fmt"
+	"os"
+)
+
+func main() {
+	server, found := os.LookupEnv("NEX_HOSTSERVICES_NATS_SERVER")
+	if !found {
+		fmt.Fprintln(os.Stderr, "failed to find NEX_HOSTSERVICES_NATS_SERVER in environment")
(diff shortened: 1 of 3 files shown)


examples/native_hostservice/main.go
```diff
@@ -0,0 +1,71 @@
...
+	})
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to subscribe to fundemo: %s\n", err)
+		return
+	}
+
+	<-ctx.Done()
+}
+
+func setupSignalHandlers(nc *nats.Conn) {
```
autodidaddict, line 52: I don't think this sample needs to show signal handler trapping, especially since this is ignored when running in a firecracker sandbox.

autodidaddict (approved):
I think we can get rid of the signal handler stuff as that is likely to distract from the actual sample.

autodidaddict (approved):

PR #337, opened 2024-08-02
stops infinite recursion. attempts to clarify difference between the 2 deploy requests

autodidaddict:
In addition to the minor fix, there's a few renames and the removal of unused properties. There is a lot more work to do on refactoring how we're using the internal agent deploy "command" and the external control deploy "request". What I've done here is just what can be done without having to rewrite any functions.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -175,7 +175,7 @@ func (a *Agent) Version() string {
 // the executable workload artifact from the cache bucket, write it to a
 // temporary file and make it executable; this method returns the full
 // path to the cached artifact if successful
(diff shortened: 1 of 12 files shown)


jordan-rash (dismissed):

kthomas:
Could we rename to AgentRequest or AgentDeployRequest instead?

autodidaddict:
> Could we rename to AgentRequest or AgentDeployRequest instead?

At the point it becomes the agent workload info, it's no longer a request. It isn't something that can be rejected for validation reasons, etc, it's a command or, as it stands now, it's a standalone piece of info.

Naming is hard.

jordan-rash (approved):

PR #339, opened 2024-08-05
Fix regression related to AgentWorkloadInfo refactor

kthomas:
#337 removed properties from the agent's version of the `DeployRequest` struct (now called `AgentWorkloadInfo` as a result of #337). This prevented essential workloads from being restarted due to missing the encrypted environment and sender public key.

### DIFF
--- a/internal/agent-api/client.go
+++ b/internal/agent-api/client.go
@@ -27,7 +27,7 @@ type ContactLostCallback func(string)
 
 const (
 	defaultAgentPingIntervalMillis = 5000
-	maxRetryAttempts               = 3
+	maxRetryAttempts               = 10
 
 	NexTriggerSubject = "x-nex-trigger-subject"
 	NexRuntimeNs      = "x-nex-runtime-ns"
--- a/internal/node/node.go
+++ b/internal/node/node.go
@@ -445,15 +445,10 @@ func (n *Node) handleAutostarts() {
 			continue
 		}
 
-		var hash string
-		if workloadHash != nil {
-			hash = *workloadHash // HACK!!! for agent-local workloads, this should be read from the release manifest
-		}
-
-		agentWorkloadInfo := agentWorkloadInfoFromControlDeployRequest(request, autostart.Namespace, numBytes, hash)
+		agentWorkloadInfo := agentWorkloadInfoFromControlDeployRequest(request, autostart.Namespace, numBytes, *workloadHash)
 
 		agentWorkloadInfo.TotalBytes = int64(numBytes)
-		agentWorkloadInfo.Hash = hash
+		agentWorkloadInfo.Hash = *workloadHash
 
 		agentWorkloadInfo.Environment = autostart.Environment // HACK!!! we need to fix autostart config to allow encrypted environment...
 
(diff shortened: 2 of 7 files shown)


internal/node/node.go
```diff
@@ -445,11 +445,19 @@ func (n *Node) handleAutostarts() {
 			continue
 		}
 
-		agentDeployRequest := agentWorkloadInfoFromControlDeployRequest(request, autostart.Namespace, numBytes, *workloadHash)
-		agentDeployRequest.TotalBytes = int64(numBytes)
-		agentDeployRequest.Hash = *workloadHash
+		var hash string
+		if workloadHash != nil {
+			hash = *workloadHash // HACK!!! for agent-local workloads, this should be read from the release manifest
```
autodidaddict, line 450: agent-local workloads don't exist(yet)

internal/agent-api/client.go
```diff
@@ -27,7 +27,7 @@ type ContactLostCallback func(string)
 
 const (
 	defaultAgentPingIntervalMillis = 5000
-	maxRetryAttempts               = 3
+	maxRetryAttempts               = 10
```
autodidaddict, line 30: does this actually need to be 10?? This seems pretty excessive and it we're waiting 10 retries for the agent, then something bigger is happening than retries.

autodidaddict (commented):
just had a question about setting retries to 10

internal/node/node.go
autodidaddict: what is a "release manifest" in this context?

internal/agent-api/client.go
kthomas: Yes... 3 wasn't enough. At 100ms intervals, it's only 1 second.

internal/node/node.go
kthomas: When we support agent-local workloads, the agent release should include a hash of the expected artifact.

internal/agent-api/client.go
kthomas: I'm not sure I agree. We want the request to succeed as soon as possible. I don't see a downside in keeping the attempt count high and the timeout at 100ms at this time.

autodidaddict:
> #337 removed properties from the agent's version of the `DeployRequest` struct (now called `AgentWorkloadInfo` as a result of #337). This prevented essential workloads from being restarted due to missing the encrypted environment and sender public key.

shouldn't the essential workload have simply stored the deployrequest rather than an agentworkloadinfo ?

kthomas:
> > #337 removed properties from the agent's version of the `DeployRequest` struct (now called `AgentWorkloadInfo` as a result of #337). This prevented essential workloads from being restarted due to missing the encrypted environment and sender public key.
> 
> shouldn't the essential workload have simply stored the deployrequest rather than an agentworkloadinfo ?

I considered that actually but it seems simpler to continue handling it the way we are. One caveat to that is: we should definitely _not_ be storing the unencrypted environment on the node. But we are doing this already. I plan to address that in a separate PR (see my comment also about the autostart workload environment also needing a mechanism to be encrypted at rest).

autodidaddict (approved):

PR #345, opened 2024-08-07
HS Config Env

jordan-rash:

### DIFF
--- a/nex/runner.go
+++ b/nex/runner.go
@@ -115,9 +115,19 @@ func RunWorkload(ctx context.Context, logger *slog.Logger) error {
 		argv = strings.Split(RunOpts.Argv, " ")
 	}
 
+	var hsConfig *controlapi.NatsJwtConnectionInfo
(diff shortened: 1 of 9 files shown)


nex/runner.go
```diff
@@ -115,9 +115,19 @@ func RunWorkload(ctx context.Context, logger *slog.Logger) error {
 		argv = strings.Split(RunOpts.Argv, " ")
 	}
 
+	var hsConfig *controlapi.NatsJwtConnectionInfo
+	if RunOpts.HsUrl != "" {
```
jordan-rash, line 119: Is this a fair assumption?  Or should all fields be present?

autodidaddict: I think using the url is good enough as an indicator for now. We can always make it.more strict later.

autodidaddict (approved):
These nested if checks on null make me miss rust options and maps.

autodidaddict (approved):

PR #349, opened 2024-08-12
fix npe

jordan-rash:


autodidaddict (approved):
:older_man: you know, none of this would be a problem in a language without nils

PR #353, opened 2024-08-14
Sandbox Default

jordan-rash:
Unfortunately, the default on the --no-sandbox flag was overriding the config everytime....this checks before it overrides

### DIFF
--- a/internal/models/cli.go
+++ b/internal/models/cli.go
@@ -132,7 +132,7 @@ type NodeOptions struct {
 	OutJSON   bool   `json:"-"`
 	NexusName string `json:"-"`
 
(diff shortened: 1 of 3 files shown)


internal/models/cli.go
```diff
@@ -132,7 +132,7 @@ type NodeOptions struct {
 	OutJSON   bool   `json:"-"`
 	NexusName string `json:"-"`
 
-	NoSandbox bool              `json:"-"`
+	NoSandbox *bool             `json:"-"`
```
autodidaddict, line 135: :sob:

autodidaddict (approved):

PR #367, opened 2024-08-25
Extracts wasm and v8 workload executors into their own separate binary plugins

### LINKED ISSUE #363: Switch from Go native plugins to external binary plugins
### Proposed change

Rather than loading dynamic libraries compiled only in Go (which therefore also prevents a plugin-enabled agent from being statically compiled), we should instead adopt the "plugins over stdio" pattern (or a similar variant), loading executable binaries and communicating with them via some transport contract. 

### Use case

This would allow plugins to be written in other languages than Go, taking advantage of different ecosystems for different workload types. Further, by no longer dynamically loading Go-based plugins, the agent wouldn't have to be statically compiled. Even further further, the agent wouldn't have to have dependencies like V8 (which has strict requirements as well as CGO issues) or WebAssembly. This leaves the agent slim, supporting only process execution for native and plugin-based workloads.

### Contribution

_No response_

autodidaddict:
This closes #363 , which switches from using Go-native plugins to simple binary plugins.


autodidaddict:
Still have a bunch to do here as well as a lot of cleanup. Just checking this into a draft PR so I don't lose it all accidentally :)

autodidaddict:
Things that still need to be done:

- [ ] make sure the rootfs stuff works by adding the `wasm` and `v8` binaries to `/usr/local/bin`
- [ ] make sure node config defaults agent plugin path to `/usr/local/bin`
- [ ] finish messaging host service wasm client
- [ ] add and finish http client host service wasm client
- [ ] add documentation Pr to docs.nats.io containing new docs on wasm type

PR #370, opened 2024-09-10
scaffolding for json schemas for agent api

autodidaddict:
initial scaffolding for an Async API definition for the communication protocol between the Nex node and the agent.

It's worth noting that the start workload request has fewer fields than does the current `AgentWorkloadInfo`. This is deliberate, because the agent doesn't need to know half of that stuff - most of that information is only needed by the host to manage the agent.


autodidaddict:
Once the asyncapi npm module is installed you can run:
```
asyncapi generate fromTemplate ./asyncapi.yaml @asyncapi/html-template -o ./docs
```
to generate html into the `docs` directory. You can use `@asyncapi/markdown-template` if you want markdown

agent-api/go/foo.go
jordan-rash, line 1: did you mean to commit `foo.go`?

jordan-rash (approved):

PR #372, opened 2024-09-12
CLI Update | `node up`

jordan-rash:


node/node.go
```diff
@@ -0,0 +1,49 @@
...
+	return nexNode, nil
+}
+
+func (nn NexNode) Validate() error {
+	var errs error
+
+	return errs
+}
+
+func (NexNode) Start() {
```
autodidaddict, line 44: should also be a pointer, and should have a name like `nn`

node/node.go
```diff
@@ -0,0 +1,49 @@
...
+	"context"
+	"fmt"
+	"log/slog"
+	"os"
+	"time"
+
+	"github.com/nats-io/nats.go"
+)
+
+type NexNode struct {
```
autodidaddict, line 13: tiny nit... but I'd like to have an interface for the node, and the node itself returns a `*nexNode` (non-exported struct)

autodidaddict (changes requested):
just a couple tiny patterns to adopt before moving on

node/node.go
```diff
@@ -10,38 +10,43 @@ import (
...
+}
+
+type nexNode struct {
 	nc *nats.Conn
 
 	logger *slog.Logger
 }
 
-type NexOption func(*NexNode)
+type NexOption func(*nexNode)
```
autodidaddict, line 24: I think this should be `func(*NodeOptions)`, and the node options are a separate struct from the nex node.

autodidaddict (changes requested):
the options builder pattern usually builds a struct separate from the target, which lets us have `NodeOptions` as an exported type and keep `nexNode` as a hidden type

node/node.go
```diff
@@ -0,0 +1,185 @@
...
+
+type nexNode struct {
+	nc *nats.Conn
+
+	logger                *slog.Logger
+	agentHandshakeTimeout int
+	resourceDirectory     string
+	internalNodeNATHost   string
+	internalNodeNATPort   int
+	microVMMode           bool
```
autodidaddict, line 31: Does the  node care about this anymore since the agent now decides whether to use firecracker?

Maybe defaultMicrovmMode or similar

jordan-rash: it cares about it for preflight/valuations....unless we are going to move that to the agents (no please).  As it sits now, we don't check or install for the firecracker/microvm dependencies unless we tell it to.  

To your point, I can see why this wouldn't be the nodes problem anymore.  That begs the question, whose job is it to check dependencies.  Example I just though of...if one day we have a "python agent" that runs scripts...whos job is it going to be to ensure the `python` binary exists?

autodidaddict:
That's probably up to the agent

node/node.go
```diff
@@ -0,0 +1,138 @@
...
+	Start()
+}
+
+type nexNode struct {
+	nc *nats.Conn
+
+	logger                *slog.Logger
+	agentHandshakeTimeout int
+	resourceDirectory     string
+	internalNodeNATHost   string
```
autodidaddict, line 28: ultra mega tiny nit ... should be `internalNodeNATSHost` and internalNodeNATSPort`
you're missing the S in NATS

cmd/nex/node.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+
+	logger.Info("Validating Nex Node")
+	err = nexNode.Validate()
+	if err != nil {
+		return err
+	}
+
+	logger.Info("Starting Nex Node")
+	nexNode.Start()
```
autodidaddict, line 284: might be worth leaving a comment here that this code requires `Start` to be blocking

autodidaddict (commented):
tiny nits

node/node.go
autodidaddict: though that internal host and port are things decided by the node... they shouldn't be configurable

autodidaddict (approved):

PR #374, opened 2024-09-13
Cleanup example config

jordan-rash:
Minimal "valid" config
```
└─❯ go run . node up --config ../../_examples/nodeconfigs/simple.json
[INFO] 2024-09-13 14:11:58 - Validating Nex Node
[INFO] 2024-09-13 14:11:58 - Starting Nex Node
[INFO] 2024-09-13 14:12:03 - Shutting down Nex Node
```


_examples/nodeconfigs/simple.json
```diff
@@ -0,0 +1,20 @@
...
+    "color": true,
+    "level": "info"
+  },
+  "nexus_name": "mynexus",
+  "tags": {
+    "foo": "bar"
+  },
+  "workload_types": [
+    {
+      "name": "CONFIGWORKLOAD",
```
autodidaddict, line 12: this isn't going to work. It will have to be something like `javascript` or `native` etc.

_examples/nodeconfigs/simple.json
```diff
@@ -0,0 +1,20 @@
...
+      "agenturi": "nats://uri/config",
```
autodidaddict, line 13: to work by default, this likely needs to be empty so the node can create a default from the resource directory + name. We can document using URIs in a different config.

autodidaddict (changes requested):
some thoughts

_examples/nodeconfigs/simple.json
jordan-rash: 🤔

_examples/nodeconfigs/simple.json
jordan-rash: This is just a "valid" config that will start the node in its current form....but will rename to native as that makes more sense

PR #375, opened 2024-09-13
initial scaffolding for ergo supervision tree

autodidaddict:

PR #377, opened 2024-09-16
initial scaffolding for ergo supervision tree

jordan-rash:
Replacing accidental close of #375


autodidaddict (dismissed):

node/node.go
```diff
@@ -22,108 +21,98 @@ type Node interface {
...
 
-	logger                *slog.Logger
-	agentHandshakeTimeout int
-	resourceDirectory     string
-	tags                  map[string]string
-	validIssuers          []string
-	otelOptions           OTelOptions
-	workloadOptions       []WorkloadOptions
-	hostServiceOptions    HostServiceOptions
+	// NOTE on refactor: the node struct for an operational node (state) should be
```
jordan-rash, line 24: I don't understand this.  Why does options need to be a separate package?  All the examples I have found using this pattern, the Options live next to the constructor.

jordan-rash (commented):
Moving options should probably be its own PR.  I don't understand why it needs it own package

node/node.go
autodidaddict: This is specifically so that the options can be a single thing that is passable around. It needs to be its own package because if the options are sitting on the `nexNode` struct, then we create an import loop.

Moving options shouldn't be its own PR because I literally can't pass options into the actors without options being in their own package.

autodidaddict:
The `models` package looks clean

PR #378, opened 2024-09-20
Make run requests more stable

kthomas:
Make run requests behave "more synchronously" and increase default agent handshake timeout.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -28,7 +28,7 @@ import (
 
 const (
 	defaultAgentHandshakeAttempts       = 5
-	defaultAgentHandshakeTimeoutMillis  = 500
+	defaultAgentHandshakeTimeoutMillis  = 5000
 	runloopSleepInterval                = 250 * time.Millisecond
 	runloopTickInterval                 = 2500 * time.Millisecond
 	workloadExecutionSleepTimeoutMillis = 50
--- a/internal/node/controlapi.go
+++ b/internal/node/controlapi.go
@@ -41,7 +41,7 @@ func NewApiListener(log *slog.Logger, mgr *WorkloadManager, node *Node) *ApiList
 	efftags := config.Tags
 	efftags[controlapi.TagOS] = runtime.GOOS
 	efftags[controlapi.TagArch] = runtime.GOARCH
-	efftags[controlapi.TagCPUs] = strconv.FormatInt(int64(runtime.NumCPU()), 10)
+	efftags[controlapi.TagCPUs] = strconv.FormatInt(int64(runtime.GOMAXPROCS(0)), 10)
 	if node.config.NoSandbox {
 		efftags[controlapi.TagUnsafe] = "true"
 	}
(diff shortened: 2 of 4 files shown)


agent/agent.go
```diff
@@ -28,7 +28,7 @@ import (
 
 const (
 	defaultAgentHandshakeAttempts       = 5
-	defaultAgentHandshakeTimeoutMillis  = 500
+	defaultAgentHandshakeTimeoutMillis  = 5000
```
bruth, line 31: How did you come to this increased timeout being more appropriate?

internal/node/controlapi.go
```diff
@@ -41,7 +41,7 @@ func NewApiListener(log *slog.Logger, mgr *WorkloadManager, node *Node) *ApiList
 	efftags := config.Tags
 	efftags[controlapi.TagOS] = runtime.GOOS
 	efftags[controlapi.TagArch] = runtime.GOARCH
-	efftags[controlapi.TagCPUs] = strconv.FormatInt(int64(runtime.NumCPU()), 10)
+	efftags[controlapi.TagCPUs] = strconv.FormatInt(int64(runtime.GOMAXPROCS(0)), 10)
```
bruth, line 44: Based on the docs, `GOMAXPROCS(0)` does not change the value that is otherwise returned by `NumCPU()` and `GOMAXPROCS(0)` will eventually be deprecated. I suggest reverting back to `NumCPU`.

internal/node/controlapi.go
```diff
@@ -421,16 +421,7 @@ func (api *ApiListener) handleDeploy(ctx context.Context, span trace.Span, m *na
 	}
 	span.AddEvent("Agent deploy request accepted")
 
-	if _, ok := api.mgr.handshakes[workloadID]; !ok {
```
bruth, line 424: This seems like a necessary check, no? There could still be a situation that the handshake doesn't occur?

nex/runner.go
```diff
@@ -55,7 +56,7 @@ func RunWorkload(ctx context.Context, logger *slog.Logger) error {
 		return err
 	}
 
-	nodeClient := controlapi.NewApiClientWithNamespace(nc, Opts.Timeout, Opts.Namespace, logger)
+	nodeClient := controlapi.NewApiClientWithNamespace(nc, time.Millisecond*25000, Opts.Namespace, logger)
```
bruth, line 59: Was it intentional to change this from an option value to a hard-coded value? Does this make the option not useable anymore? And again, what led to determining this specific timeout?

bruth (commented):
Some questions and one comment. It would be helpful to understand the correlation of these changes to how it makes the run request more stable. Is there a test that can be written to demonstrate this before and after?

internal/node/controlapi.go
autodidaddict: This actually doesn't change anything meaningful, this is only changing the value of the CPU tag that the node reports upon query.

kthomas: Using `runtime.NumCPU()` is actually a bug-- it returns the number "available" on the machine, not the number `GOMAXPROCS`. Using `runtime.GOMAXPROCS(0)` actually returns the number of CPUs configured via `GOMAXPROCS`.

nex/runner.go
kthomas: In order to make the run requests behave synchronously, I changed this to `25000` milliseconds to match the maximum amount of time `MachineManager` would wait for an agent to be available (which is now 5 attempts * 5000ms).

agent/agent.go
kthomas: Given that `500` milliseconds seemed sufficient, `5000` should be more than sufficient. I hard-coded a reduced number of CPUs available to the nex node to simulate the resources made available in GCP during testing.

internal/node/controlapi.go
kthomas: We could alternatively parse it out of `os.Getenv`.

jordan-rash: > We could alternatively parse it out of os.Getenv

This will only work if GOMAXPROCS is overridden by environment

kthomas: For some reason I was thinking the process set `GOMAXPROCS` if otherwise unset. But that makes sense.

nex/runner.go
kthomas: > (which is now 5 attempts * 5000ms).

Actually, this is only true now as a result of [this](https://github.com/synadia-io/nex/pull/378/commits/3355cfd8993692ec35c4233222e5e11b602c9c6d) commit.

PR #379, opened 2024-09-20
Refactor to support "agent manager" pattern

kthomas:
This PR introduces a way to support management of agents per workload type.


bruth:
@kthomas I know this is still in draft, but can you please add comments to the new code as well as additional tests?

PR #381, opened 2024-09-24
CLI rootfs Subcommand

jordan-rash:
To build the equivalent to the rootfs today 
```
sudo -E go run . rootfs docker.io/synadia/nex-rootfs:alpine --files ./nex-agent_0.3.0-rc1_linux_amd64=/usr/local/bin/agent
```


kthomas:
The [umoci](https://github.com/opencontainers/umoci) project has an [unpack](https://github.com/opencontainers/umoci/blob/main/unpack.go#L35) method that was being used successfully in the OCI spike branch. Just sharing it here for posterity, the way you have implemented it here looks good.

cmd/nex/rootfs.go
```diff
@@ -1,3 +1,389 @@
...
+		repo.PlainHTTP = true
+	}
+	r.ociRef.Repo = repo
+
+	repo.Client = &auth.Client{
+		Cache: auth.DefaultCache,
+		Credential: func(ctx context.Context, registry string) (auth.Credential, error) {
+			switch r.ociRef.Repo.Reference.Registry {
+			case "docker.io":
+				authGet, err := http.Get(fmt.Sprintf("https://auth.docker.io/token?service=registry.docker.io&scope=repository:%s:pull", repo.Reference.Repository))
```
kthomas, line 100: Helper function to return the URL might be nice.

cmd/nex/rootfs.go
```diff
@@ -1,3 +1,389 @@
...
+
+	if r.ociRef.Manifest == nil {
+		return errors.New("no matching manifest found")
+	}
+
+	return nil
+}
+
+func (r *RootFS) Validate() error {
+	return nil
```
kthomas, line 178: How will validation work?

jordan-rash: Kong runs the Validate function on all commands automatically, so by satisfying the interface, we will get validation for free.  That said, there are no validations from rootfs command yet

cmd/nex/rootfs.go
jordan-rash: I decided against it because it's only for docker.io and the URL is always the same.  I can pull it up though to make it easier to read

cmd/nex/rootfs.go
jordan-rash: Just added them

cmd/nex/rootfs_test.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func TestRootFsSimple(t *testing.T) {
+	if os.Geteuid() != 0 {
```
jordan-rash, line 12: Need to add CI to be able to run this test (sudo)

autodidaddict (approved):

PR #389, opened 2024-09-30
Improve scaffolding

kthomas:
This PR improves on the initial scaffolding for the node.


node/actors/app.go
```diff
@@ -2,30 +2,41 @@ package actors
 
 import (
 	"ergo.services/ergo/gen"
+	"github.com/nats-io/nats.go"
 	"github.com/synadia-io/nex/models"
 )
 
+const nexApplicationName = "nex"
```
jordan-rash, line 9: please 
```
const(
...
) 
```

node/actors/control_api.go
```diff
@@ -1,26 +1,197 @@
...
+	}
+
+	api.publicXKey = publicXKey
+	return nil
+}
+
+func (api *controlAPI) shutdown() error {
+	var err error
+
+	for _, sub := range api.subsz {
```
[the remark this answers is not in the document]
autodidaddict: This was a bad pattern before and we shouldn't copy it over. Call `Drain` once on the connection.

node/actors/internal_nats_server.go
```diff
@@ -62,9 +69,9 @@ func (ns *internalNatsServer) HandleMessage(from gen.PID, message any) error {
 			return err
 		}
 		ns.tokens[InternalNatsServerReadyName] = token
-		ns.Log().Info("registered publishable event %s, waiting for consumers...", InternalNatsServerReadyName)
+		ns.Log().Info("registered publishable event, waiting for consumers...", slog.Any("atom", InternalNatsServerReadyName))
```
autodidaddict, line 72: the field name here isn't `atom`, it's the name of a published event, so call it `event_name`

node/actors/internal_nats_server.go
```diff
@@ -62,9 +69,9 @@ func (ns *internalNatsServer) HandleMessage(from gen.PID, message any) error {
...
 	case eventStart:
-		ns.Log().Info("publisher got first consumer for %s. start producing events...", InternalNatsServerReadyName)
+		ns.Log().Info("publisher got first consumer. start producing events...", slog.Any("atom", InternalNatsServerReadyName))
```
autodidaddict, line 74: do not use `atom` here, that's meaningless to someone viewing the logs. it's an `event_name`

node/actors/logger.go
```diff
@@ -26,7 +26,20 @@ type NodeLogger struct {
 func (nl *NodeLogger) HandleLog(message gen.MessageLog) error {
 	// TODO: this is temporary until we can figure out what we really want
 	// to do with these logs
-	nl.inner.Log(context.Background(), slog.Level(message.Level), fmt.Sprintf(message.Format, message.Args...))
+	switch message.Level {
```
autodidaddict, line 29: why is this switch statement necessary when the call to `Log` did the job?

autodidaddict:
In order to avoid a firehose of GitHub notifications, can you leave this in draft until you want it reviewed?

node/actors/logger.go
jordan-rash: The log message doesn't do the job.  Here's why ->
From ergo:
```
const (
	LogLevelSystem LogLevel = -100

	LogLevelTrace    LogLevel = -2
	LogLevelDebug    LogLevel = -1
	LogLevelDefault  LogLevel = 0 // inherite from node/app/parent process
	LogLevelInfo     LogLevel = 1
	LogLevelWarning  LogLevel = 2
	LogLevelError    LogLevel = 3
	LogLevelPanic    LogLevel = 4
	LogLevelDisabled LogLevel = 5
)
```
From slog: [link](https://pkg.go.dev/log/slog#Level)
```
const (
	LevelDebug Level = -4
	LevelInfo  Level = 0
	LevelWarn  Level = 4
	LevelError Level = 8
)
```

As you can see, the mappings are incorrect.  The ergo author either rolled his own or used something not IAW the go standard lib

node/actors/internal_nats_server.go
kthomas: Copypasta artifact.

node/actors/internal_nats_server.go
kthomas: Copypasta artifact.

node/actors/logger.go
jordan-rash: Logs without the translation: 
```
[INFO] 2024-09-30 14:05:49 - Validating Nex Node
[INFO] 2024-09-30 14:05:49 - Starting Nex Node
[] 2024-09-30 14:05:49 - Nex node started
[] 2024-09-30 14:05:49 - Observer Application started and available at http://localhost:9911/
[] 2024-09-30 14:05:49 - registered publishable event 'internal_nats_server_ready', waiting for consumers...
[] 2024-09-30 14:05:49 - successfully subscribed to: Event#<EC87E8E8:'internal_nats_server_ready'>
[] 2024-09-30 14:05:49 - publisher got first consumer for 'internal_nats_server_ready'. start producing events...
[] 2024-09-30 14:05:49 - received event Event#<EC87E8E8:'internal_nats_server_ready'>
[] 2024-09-30 14:05:49 - starting 0 agent binaries..
``` 
(missing level information 

Logs after fix in place 
```
[INFO] 2024-09-30 14:26:46 - Validating Nex Node
[INFO] 2024-09-30 14:26:46 - Starting Nex Node key=derp
[INFO] 2024-09-30 14:26:46 - Nex node started%!(EXTRA slog.Attr=sadf=Asdf)
[INFO] 2024-09-30 14:26:46 - Observer Application started and available at http://localhost:9911/
[INFO] 2024-09-30 14:26:46 - registered publishable event 'internal_nats_server_ready', waiting for consumers...
[INFO] 2024-09-30 14:26:46 - successfully subscribed to: Event#<EC87E8E8:'internal_nats_server_ready'>
[INFO] 2024-09-30 14:26:46 - publisher got first consumer for 'internal_nats_server_ready'. start producing events...
[INFO] 2024-09-30 14:26:46 - received event Event#<EC87E8E8:'internal_nats_server_ready'>
[INFO] 2024-09-30 14:26:46 - starting 0 agent binaries..
```

node/actors/control_api.go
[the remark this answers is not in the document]
kthomas: If we don't drain these subscriptions and the actor restarts, we will end up with duplicate subscriptions. The underlying NATS connection is used for other things in the node so we shouldn't call `Drain` on it from the control API-- we do need to call `Drain` on it though during node shutdown and I will add that to this PR.

Also, it is unclear when the `controlAPI` actor would actually restart, since there isn't anything actually running in the background. We should clarify this more broadly when it comes to some of the other actors as well that seemingly will never exit unexpectedly.

kthomas:
> Lets leave this as a draft until you're ready for it to be reviewed

Just [this](https://github.com/synadia-io/nex/pull/389#discussion_r1781970270) item remains pending.

node/actors/logger.go
kthomas: Thanks @jordan-rash.

Marking this as resolved. @autodidaddict please unresolve if your concern wasn't addressed.

node/actors/control_api.go
```diff
@@ -1,26 +1,197 @@
...
 	act.Actor
+
+	publicKey  string
+	publicXKey string
+	subsz      []*nats.Subscription
 }
 
-func (api *controlAPIServer) Init(args ...any) error {
-	api.Log().Info("Control API Server started")
+func (api *controlAPI) Init(args ...any) error {
```
bruth, line 30: For all of these `Init` methods that take variable arguments, it would be safer and easier to maintain/evolve if a single struct was passed with the arguments rather than relying on them positionally. Only one argument would need to be type-checked and validation could be more flexible.

node/actors/control_api.go
```diff
@@ -1,26 +1,197 @@
...
+	"github.com/nats-io/nkeys"
 )
 
-func createControlAPIServer() gen.ProcessBehavior {
-	return &controlAPIServer{}
+const APIPrefix = "$NEX"
+
+func createControlAPI() gen.ProcessBehavior {
+	return &controlAPI{
+		subsz: make([]*nats.Subscription, 0),
```
bruth, line 18: No need to initialize a slice explicitly if a length or cap is not required.

node/actors/control_api.go
```diff
@@ -1,26 +1,197 @@
...
-func (api *controlAPIServer) HandleInspect(from gen.PID, item ...string) map[string]string {
+func (api *controlAPI) HandleInspect(from gen.PID, item ...string) map[string]string {
 	api.Log().Info("control API server got inspect request from %s", from)
 	return nil
 }
+
+func (api *controlAPI) initPublicXKey() error {
+	kp, err := nkeys.CreateCurveKeys()
+	if err != nil {
+		api.Log().Error("Failed to create x509 curve key", slog.Any("error", err))
```
bruth, line 75: xkeys are not x.509

kthomas: Good catch... this is an artifact from copying code over from main.

node/actors/control_api.go
autodidaddict: This makes a lot of sense. Each actor can expose a struct that contains the init options it needs and then just get it from args[0]

Note that the args slice is how ergo inits stuff and there's a whole rabbit hole behind why. (It's related to failover restarts)

node/actors/agent_manager.go
```diff
@@ -37,18 +39,18 @@ func (mgr *agentManager) HandleInspect(from gen.PID, item ...string) map[string]
...
 		}
-		mgr.Log().Info("successfully subscribed to: %s", InternalNatsServerReady)
+		mgr.Log().Info("successfully subscribed to internal nats server", slog.Any("event_name", InternalNatsServerReady))
 	}
 	return nil
 }
 
 func (mgr *agentManager) HandleEvent(event gen.MessageEvent) error {
-	mgr.Log().Info("received event %s", event.Event)
+	mgr.Log().Info("received event", slog.Any("event", event.Event))
```
autodidaddict, line 53: we should be consistent here. if we log event names as `event_name` (line 47 above) then we should always use that field to make log searches easier.

node/actors/control_api.go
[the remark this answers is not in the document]
autodidaddict: the actor restart policy is declared explicitly by their supervisor. Manually invoking `shutdown` like this implies that you're shutting it all down and that stuff isn't going to come back... which means we can do a connection-wide drain.

If we were only draining the stuff for controlAPI because we thought it was coming back up, then the subsz makes sense. But, as written this kind of shutdown is permanent.

tl;dr:
* if the shutdown is going to be a restart, use subsz (doesn't look like this is the case)
* if the shutdown is permanent, use connection-wide single-call Drain

node/actors/supervisor.go
```diff
@@ -18,27 +29,55 @@ type NexSupervisor struct {
 func (sup *NexSupervisor) Init(args ...any) (act.SupervisorSpec, error) {
 	var spec act.SupervisorSpec
 
-	nodeOptions := args[0].(models.NodeOptions)
+	if len(args) != 3 {
```
autodidaddict, line 32: see @bruth 's previous comment. Things get a lot cleaner if we pass a single struct that holds the node id, the connection, and the node options.

node/node.go
```diff
@@ -120,17 +121,25 @@ func (nn *nexNode) Validate() error {
...
-	nn.initializeSupervisionTree()
+func (nn *nexNode) Start() error {
+	return nn.initializeSupervisionTree()
 }
 
-func (nn *nexNode) initializeSupervisionTree() {
+func (nn *nexNode) initializeSupervisionTree() error {
 	var options gen.NodeOptions
 
+	nodeName := "nex@localhost" // FIXME-- make configurable
```
autodidaddict, line 131: I actually think this shouldn't be configurable. This is only used for ergo internals that make use of a clustering feature we don't use.

autodidaddict (changes requested):

node/actors/agent_manager.go
kthomas: I noticed this also. Will fix.

node/actors/control_api.go
bruth: I figured it was an interface requirement.

autodidaddict: Yep. Also, the Ergo runtime keeps a copy of those args in case an actor crashes and it needs to restart it. The constructors are never re-invoked, but the "factory" method is called and then the `Init` function is re-invoked with the original params.

kthomas: All actors on this branch with the exception of `internalNatsServer` now conform to this suggested single-argument `params`. Will add the same for the internal NATS actor as part of #390 after we merge this PR.

node/node.go
```diff
@@ -120,45 +127,79 @@ func (nn *nexNode) Validate() error {
...
 	options.Log.DefaultLogger.Disable = true
 
 	// https://docs.ergo.services/basics/logging#process-logger
 	// https://docs.ergo.services/tools/observer#log-process-page
 
-	nodeName := "nex@localhost"
-
 	// starting node
-	node, err := ergo.StartNode(gen.Atom(nodeName), options)
+	node, err := ergo.StartNode(gen.Atom(nn.name), options)
```
autodidaddict, line 156: I specifically indicated that this needs to be a constant that is not mutable or configurable.

autodidaddict (changes requested):

node/actors/control_api.go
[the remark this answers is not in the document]
kthomas: This makes sense, @autodidaddict.

In the meantime, can you please address the note related to how some of the defined actors will actually never exit unexpectedly? (e.g., the internal NATS server, control API, and host services actors should never actually "crash")

We are still using the `Actor` pattern for these structs. Was the rational primarily to support ergo observability for these?

node/node.go
kthomas: It is not configurable: https://github.com/synadia-io/nex/blob/feat/vnext-node/node/node.go#L42

This seemed like a good compromise since we log the name quite a bit.

node/actors/control_api.go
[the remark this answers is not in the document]
autodidaddict: If one of those actors has a panic, ergo will halt that actor. Observability is like a free bonus. The main reason for ergo is so that we don't have to write a single mutex in the code anywhere.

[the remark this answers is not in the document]
autodidaddict: After thinking about this during lunch, the `subsz` pattern is probably safer, even if we don't expect to be bouncing those actors. Carry on :) 
image

node/node.go
autodidaddict: ok. I saw the `default` prefix and assumed there was a way to override it. Let's change the constant to something like `ergoNodeName` so that it's clear it's not actually related to the "nex node".

autodidaddict: We also shouldn't be logging that name at all. It's not useful for anything. (Unless it's coming "for free" with the actor logging, in which case we can maybe tweak that so we don't log the node name in another PR)

autodidaddict (approved):

PR #390, opened 2024-09-30
Implement internal NATS actor

kthomas:


node/actors/internal_nats_server.go
```diff
@@ -25,23 +45,64 @@ type agentCredential struct {
...
+		return err
+	}
+
+	err = ns.startNatsServer(opts)
 	if err != nil {
 		return err
 	}
-	_ = ns.Send(ns.PID(), "post_init")
+
+	err = ns.Send(ns.PID(), "post_init")
```
autodidaddict, line 100: we should probably adopt the pattern from the other PR and remove this magic string and replace with a constant `PostInit`

node/actors/internal_nats_server.go
```diff
@@ -62,9 +123,9 @@ func (ns *internalNatsServer) HandleMessage(from gen.PID, message any) error {
 			return err
 		}
 		ns.tokens[InternalNatsServerReadyName] = token
-		ns.Log().Info("registered publishable event %s, waiting for consumers...", InternalNatsServerReadyName)
+		ns.Log().Info("registered publishable event, waiting for consumers...", slog.Any("atom", InternalNatsServerReadyName))
```
autodidaddict, line 126: this is not an atom from a logging standpoint. This is an `event_name`

node/actors/internal_nats_server.go
```diff
@@ -62,9 +123,9 @@ func (ns *internalNatsServer) HandleMessage(from gen.PID, message any) error {
...
 	case eventStart:
-		ns.Log().Info("publisher got first consumer for %s. start producing events...", InternalNatsServerReadyName)
+		ns.Log().Info("publisher got first consumer. start producing events...", slog.Any("atom", InternalNatsServerReadyName))
```
autodidaddict, line 128: change "atom" to "event_name"

node/actors/internal_nats_server.go
```diff
@@ -95,8 +156,182 @@ func (ns *internalNatsServer) buildAgentCredentials() ([]agentCredential, error)
...
+
+	t := template.Must(template.New("natsconfig").Parse(configTemplate))
+	err := t.Execute(&wr, config)
+	if err != nil {
+		return nil, err
+	}
+
+	// -8 is equivalent to TRACE-ish level
+	// FIXME... .Log() does not exist
+	// ns.Log().Log(context.Background(), -8, "generated NATS config", ns.Log().String("config", wr.String()))
```
autodidaddict, line 262: ns.Log() does indeed exist. We should be able to use ns.Log().LevelFunc(...,) here.

autodidaddict: but we definitely don't need to log the whole config file output.

autodidaddict (changes requested):

node/actors/internal_nats_server.go
```diff
@@ -25,23 +45,64 @@ type agentCredential struct {
...
 }
 
+type configTemplateData struct {
+	Credentials       map[string]*credentials
+	Connections       map[string]*nats.Conn
+	NexHostUserPublic string
+	NexHostUserSeed   string
+}
+
+type credentials struct {
```
kthomas, line 55: I still think this is redundant and we should consider reusing `agentCredential`.

node/actors/internal_nats_server.go
kthomas: This was accidental.

node/actors/internal_nats_server.go
kthomas: It doesn't map nicely to slog levels. We have a fix for this in #389.

node/actors/internal_nats_server.go
autodidaddict: I still disagree. I don't like the idea of reusing a data structure for internal functional code for stuff used solely as the data source for a template generator. It might seem like a nitpick, but I don't like bleeding responsibilities like this. If we combine, then one purpose will make no use of two fields and the other purpose will make no use of one field. Keeping them separate means:
* it's clear that these data structures are meant for different purposes
* if we want to add additional stuff to the template generating struct later, it's clear how to do that and we don't have to disentangle them.
* we don't have the dangling orphan fields. While it might be obvious to us now what those orphan fields are for now, us 6 months from now will be confused by this reuse and entanglement.

node/actors/internal_nats_server.go
kthomas: See #389.

node/actors/internal_nats_server.go
autodidaddict: it mapped identically before. I was able to use the slog level as input. Instead of making a big switch statement and then making explicit `Error` and `Info` calls, why not create a simple mapping function that converts the ergo log level to a slog log level?

node/actors/internal_nats_server.go
kthomas: It's package-private so it seemed a little easier to read. Just wanted to note this here for posterity 👍🏻

node/actors/internal_nats_server.go
kthomas: Can you add this feedback to #389?

autodidaddict (approved):

autodidaddict (approved):

PR #408, opened 2024-10-27
adds a sample actor interaction for querying the list of running workloads from the info endpoint

autodidaddict:
Note that there is some implementation stuff that hasn't happened yet because neither of the agent actors are currently  maintaining their workload list. Once they do that, we can fill in the "make this real" bits with the actual list of workloads.

### DIFF
--- a/node/actors/control_api.go
+++ b/node/actors/control_api.go
@@ -269,6 +269,6 @@ func respondEnvelope(m *nats.Msg, dataType string, code int, data interface{}, e
 	if len(strings.TrimSpace(err)) == 0 {
 		e = nil
 	}
-	bytes, _ := json.Marshal(newEnvelope(dataType, data, code, &err))
+	bytes, _ := json.Marshal(newEnvelope(dataType, data, code, e))
 	_ = m.Respond(bytes)
 }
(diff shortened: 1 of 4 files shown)


node/actors/control_api.go
```diff
@@ -189,3 +239,36 @@ func (api *ControlAPI) handleUndeploy(m *nats.Msg) {
...
+
+func failEnvelope(dataType string, code int, err string) Envelope {
+	return Envelope{
+		PayloadType: dataType,
+		Code:        code,
+		Error:       &err,
+	}
+}
+
+func respondFail(m *nats.Msg, dataType string, code int, err string) {
```
jordan-rash, line 266: I dont love having these separate envelope functions.  Can we condense them to a `responseEnvelope(dataType string, code int, response any, err error)` and just have the one function?  `Data` and `Error` are already `omitempty` fields, so this just simplifies/unifies some things

autodidaddict: I like having them named like this because:

```
respondFail(m, MyType, 404, e)
```
is easier to quickly grasp than:
```
respondEnvelope(m, MyType, 200, nil)
```

I guess if we stick to html response code conventions it isn't that much more difficult to read.

jordan-rash (approved):

PR #410, opened 2024-10-28
NodeControl Client 

jordan-rash:
Not sure that all these fields are needed, but this is just a starting point for us to iterate on


api/nodecontrol/client.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func (c *ControlAPIClient) DeployWorkload(nodeId string) (*nodegen.StartWorkloadResponseJson, error) {
+	req := nodegen.StartWorkloadRequestJson{}
+	req_b, err := json.Marshal(req)
+	if err != nil {
+		return nil, err
+	}
+
+	msg, err := c.nc.Request(actors.DeploySubject(nodeId), req_b, DefaultRequestTimeout)
```
autodidaddict, line 97: the subject you get back from this will have a `*` in place where the namespace needs to be. We can't make requests on wildcard subjects.  There's a difficulty in using the subscription subject functions (which include wildcards) as the request subject functions.

e.g. we should have a `DeploySubscribeSubject(nodeId)` and then `DeployRequestSubject(namespace, nodeId)`

api/nodecontrol/client.go
```diff
@@ -0,0 +1,160 @@
...
+	err = json.Unmarshal(msg.Data, resp)
+	if err != nil {
+		return nil, err
+	}
+
+	return resp, nil
+}
+
+func (c *ControlAPIClient) GetInfo(nodeId string) (*nodegen.NodeInfoResponseJson, error) {
+	msg, err := c.nc.Request(actors.InfoSubject(nodeId), nil, DefaultRequestTimeout)
```
autodidaddict, line 133: the info subject for requests needs to contain a namespace (so that the node can filter the payload by namespace). So maybe `InfoSubscribeSubject(nodeId)` and `InfoRequestSubject(namespace, nodeId)`

api/nodecontrol/client.go
```diff
@@ -0,0 +1,160 @@
...
+	err = json.Unmarshal(msg.Data, resp)
+	if err != nil {
+		return nil, err
+	}
+
+	return resp, nil
+}
+
+func (c *ControlAPIClient) FindAgent() (*nodegen.AgentPingResponseJson, error) {
+	msg, err := c.nc.Request(actors.AgentPingSubject(), nil, DefaultRequestTimeout)
```
autodidaddict, line 62: we can't use `AgentPingSubject()` here because that is defined as `AgentPingSubject NexSubject = func(_ ...string) string { return APIPrefix + ".APING.>"}`

You can't make a request on `.APING.>`

When making a request, this can be `APING.{namespace}` (which retrieves all agents in the namespace), or it can be `APING.{namespace}.{workload}` where `workload` can be the workload name (potentially returning multiple results) or the workload Id, which will return 1 or 0 results.

autodidaddict (changes requested):
the client won't work on the current set of subjects, left comments detailing what should change

autodidaddict (approved):

PR #413, opened 2024-10-29
Conversions and Handlers

jordan-rash:


node/actors/pb/agents.proto
```diff
@@ -17,39 +17,80 @@ message Ping {
...
+  repeated string argv = 1;
+  string description = 2;
+  string environment = 3;
+  bool essential = 4;
+  string hash = 5;
+  HostServicesConfig host_service_config = 6;
+  string jsdomain = 7;
+  string retried_at = 8;
+  int32 retry_count = 9;
+  string sender_public_key = 10;
```
autodidaddict, line 29: we don't need the sender key, or the target node, etc. Keep in mind that the protos are for internal use and so we can remove any fields we no longer need after the API server has validated them.

jordan-rash: yea, I was just copying a 1:1 over and was going to audit after the fact, but ill pull these two out

node/actors/control_api.go
```diff
@@ -181,16 +181,108 @@ func (api *ControlAPI) subscribe() error {
...
+
+	ctx := context.Background()
+	_, agentSuper, err := api.self.ActorSystem().ActorOf(ctx, AgentSupervisorActorName)
+	if err != nil {
+		api.logger.Error("Failed to locate agent supervisor actor", slog.Any("error", err))
+		respondEnvelope(m, AuctionResponseType, 500, nil, fmt.Sprintf("failed to locate agent supervisor actor: %s", err))
+		return
+	}
+
+	askResp, err := api.self.Ask(ctx, agentSuper, auctionRequestToProto(req))
```
autodidaddict, line 200: the auction request may not need to go to the agent supervisor. Can the auction be satisfied with just node config and whatever info the api server has on hand?

jordan-rash: yea, this was my confusion.  Seems like this (and the few other node specific requests) Ask needs to go to the node supervisor I would think?

node/actors/control_api.go
```diff
@@ -199,44 +291,107 @@ func (api *ControlAPI) handleInfo(m *nats.Msg) {
 		return
 	}
 	// Ask the agent supervisor for a list of all the workloads from all of its children
-	response, err := api.self.Ask(ctx, agentSuper, req)
+	response, err := api.self.Ask(ctx, agentSuper, new(actorproto.GetNodeInfo))
 	if err != nil {
-		api.logger.Error("Failed to get list of running workloads from agent supervisor", slog.Any("error", err))
-		respondEnvelope(m, InfoResponseType, 500, nil, fmt.Sprintf("failed to get list of running workloads: %s", err))
+		api.logger.Error("Failed to get node info", slog.Any("error", err))
```
autodidaddict, line 296: This actually made more sense the other way. We're asking the agent supervisor for the list of agents and their running workloads. We use that information as _part_ of the node info, but the agent supervisor shouldn't be handling a "node info" query

node/actors/control_api.go
```diff
@@ -199,44 +291,107 @@ func (api *ControlAPI) handleInfo(m *nats.Msg) {
...
 func (api *ControlAPI) handleLameDuck(m *nats.Msg) {
-	// TODO
+	ctx := context.Background()
+	_, agentSuper, err := api.self.ActorSystem().ActorOf(ctx, AgentSupervisorActorName)
+	if err != nil {
+		api.logger.Error("Failed to locate agent supervisor actor", slog.Any("error", err))
+		respondEnvelope(m, LameDuckResponseType, 500, nil, fmt.Sprintf("failed to locate agent supervisor actor: %s", err))
+		return
+	}
+	// Ask the agent supervisor for a list of all the workloads from all of its children
```
autodidaddict, line 319: copy/paste fail

node/actors/control_api.go
```diff
@@ -199,44 +291,107 @@ func (api *ControlAPI) handleInfo(m *nats.Msg) {
...
+	response, err := api.self.Ask(ctx, agentSuper, new(actorproto.SetLameDuck))
+	if err != nil {
+		api.logger.Error("Failed to put node in lame duck mode", slog.Any("error", err))
+		respondEnvelope(m, LameDuckResponseType, 500, nil, fmt.Sprintf("failed to put node in lame duck mode: %s", err))
+		return
+	}
+
+	workloadResponse, ok := response.(*actorproto.LameDuckResponse)
+	if !ok {
+		api.logger.Error("Workload listing response from agent supervisor was not the correct type")
```
autodidaddict, line 329: wrong comment in the text (copy/paste fail)

node/actors/control_api.go
```diff
@@ -199,44 +291,107 @@ func (api *ControlAPI) handleInfo(m *nats.Msg) {
...
+	_, agentSuper, err := api.self.ActorSystem().ActorOf(ctx, AgentSupervisorActorName)
+	if err != nil {
+		api.logger.Error("Failed to locate agent supervisor actor", slog.Any("error", err))
+		respondEnvelope(m, LameDuckResponseType, 500, nil, fmt.Sprintf("failed to locate agent supervisor actor: %s", err))
+		return
+	}
 
-func (api *ControlAPI) handleUndeploy(m *nats.Msg) {
-	// TODO
+	response, err := api.self.Ask(ctx, agentSuper, new(actorproto.PingNode))
```
autodidaddict, line 346: pingnode shouldn't be something we ask the agent supervisor for. The control API should return a ping response, which may contain information obtained from the agent supervisor. Ideally it's probably coming from the same "Get workloads" query that the agent supervisor handles for node info.

node/actors/control_api.go
```diff
@@ -199,44 +291,107 @@ func (api *ControlAPI) handleInfo(m *nats.Msg) {
...
+	response, err := api.self.Ask(ctx, agentSuper, new(actorproto.PingAgent))
+	if err != nil {
+		api.logger.Error("Failed to ping agent", slog.Any("error", err))
+		respondEnvelope(m, AgentPingResponseType, 500, nil, fmt.Sprintf("failed to ping agent: %s", err))
+		return
+	}
+
+	workloadResponse, ok := response.(*actorproto.PingAgentResponse)
+	if !ok {
+		api.logger.Error("Workload listing response from agent supervisor was not the correct type")
```
autodidaddict, line 389: wrong error text (copy/paste fail)

autodidaddict (changes requested):
See individual comments. There are some copy/paste induced typos, but also a spot here and there where I think the responsibility and naming needs to be tweaked

node/actors/control_api.go
autodidaddict: There is no node supervisor ;) We should be able to make a regular function call to get the data necessary to respond to these queries.

node/actors/control_api.go
```diff
@@ -181,16 +181,108 @@ func (api *ControlAPI) subscribe() error {
...
-	// TODO
+	req := new(nodecontrol.StartWorkloadRequestJson)
+	err := json.Unmarshal(m.Data, req)
+	if err != nil {
+		api.logger.Error("Failed to unmarshal deploy request", slog.Any("error", err))
+		respondEnvelope(m, RunResponseType, 500, nil, fmt.Sprintf("failed to unmarshal deploy request: %s", err))
+		return
+	}
+
+	ctx := context.Background()
```
autodidaddict, line 226: here we don't need to talk to the agent supervisor. Since we have the workload type name on the deployment request, we can use `actorOf` to find the agent by name, and send the deploy request to it. The agent supervisor doesn't need to proxy that message.

autodidaddict (commented):
Added another comment

node/actors/control_api.go
jordan-rash: yup...not even sure why I changed it

node/actors/control_api.go
```diff
@@ -181,62 +180,198 @@ func (api *ControlAPI) subscribe() error {
...
+
+	auctResp, err := api.auctionResponse(string(req.Os), string(req.Arch), convertedAgentType, req.Tags.Tags)
+	if err != nil {
+		api.logger.Error("Failed to generate auction response", slog.Any("error", err))
+		respondEnvelope(m, AuctionResponseType, 500, nil, fmt.Sprintf("failed to generate auction response: %s", err))
+		return
+	}
+
+	if auctResp == nil {
+		respondEnvelope(m, AuctionResponseType, 404, nil, "node does not satisfy auction request")
```
autodidaddict, line 204: if we don't pass the auction, we need to not respond to the auction at all. In a scatter/gather like this, we can't respond negatively.

node/actors/control_api.go
```diff
@@ -181,62 +180,198 @@ func (api *ControlAPI) subscribe() error {
...
+	err := json.Unmarshal(m.Data, req)
+	if err != nil {
+		api.logger.Error("Failed to unmarshal undeploy request", slog.Any("error", err))
+		respondEnvelope(m, StopResponseType, 500, nil, fmt.Sprintf("failed to unmarshal undeploy request: %s", err))
+		return
+	}
 
-	req := &actorproto.QueryWorkloads{}
+	ctx := context.Background()
+	_, agentSuper, err := api.self.ActorSystem().ActorOf(ctx, AgentSupervisorActorName)
```
autodidaddict, line 255: we should be sending the undeploy request directly to the agent, the same way we send the deploy request. The agent super should only be involved in operations that involve treating the list of agents as an aggregate or starting/stopping agents.

autodidaddict: Remember that undeploy here means undeploy a workload, not stop the agent.

node/actors/control_api.go
```diff
@@ -181,62 +180,198 @@ func (api *ControlAPI) subscribe() error {
...
 	_, agentSuper, err := api.self.ActorSystem().ActorOf(ctx, AgentSupervisorActorName)
 	if err != nil {
 		api.logger.Error("Failed to locate agent supervisor actor", slog.Any("error", err))
 		respondEnvelope(m, InfoResponseType, 500, nil, fmt.Sprintf("failed to locate agent supervisor actor: %s", err))
 		return
 	}
+
 	// Ask the agent supervisor for a list of all the workloads from all of its children
-	response, err := api.self.Ask(ctx, agentSuper, req)
+	response, err := api.self.Ask(ctx, agentSuper, new(actorproto.GetNodeInfo))
```
autodidaddict, line 289: agent super shouldn't respond to get node info... we should return a node info from inside this control api, and only ask the agent super for its list of workloads.

Information like the node Id, target xkey, etc, needs to come from node configuration on the api struct and not the agent super.

node/actors/control_api.go
```diff
@@ -181,62 +180,198 @@ func (api *ControlAPI) subscribe() error {
...
+	req := new(nodecontrol.AgentPingResponseJson)
+	err := json.Unmarshal(m.Data, req)
+	if err != nil {
+		api.logger.Error("Failed to unmarshal agent ping request", slog.Any("error", err))
+		respondEnvelope(m, AgentPingResponseType, 500, nil, fmt.Sprintf("failed to unmarshal agent ping request: %s", err))
+		return
+	}
+
+	ctx := context.Background()
+	_, agentSuper, err := api.self.ActorSystem().ActorOf(ctx, AgentSupervisorActorName)
```
autodidaddict, line 353: shouldn't we be pulling the agent name/workload type from the `APING` subject, and pinging the specific agent? If that's the case, we should locate the individual agent process and not the agent super. We should only ping the agent super if we want it to aggregate ping responses from all agents.

autodidaddict (changes requested):
just need to change a few actor targets

node/node.go
```diff
@@ -181,18 +202,22 @@ func (nn *nexNode) initializeSupervisionTree() (goakt.ActorSystem, error) {
 	if err != nil {
 		return nil, err
 	}
-	inats := actors.CreateInternalNatsServer(*nn.options)
+	nn.runningActors[actors.AgentSupervisorActorName] = agentSuper
```
autodidaddict, line 205: we definitely shouldn't have to maintain a map of actors. We can look up any actor by its name, and we can obtain its list of children with the `Children` function.

node/node.go
```diff
@@ -27,8 +35,10 @@ type nexNode struct {
 	ctx context.Context
 	nc  *nats.Conn
 
-	options   *models.NodeOptions
-	publicKey nkeys.KeyPair
+	options       *models.NodeOptions
+	publicKey     nkeys.KeyPair
+	startedAt     time.Time
+	runningActors map[string]*goakt.PID
```
autodidaddict, line 41: the actor system is doing this for us. We shouldn't have to maintain this.

node/node.go
```diff
@@ -212,10 +237,13 @@ func (nn *nexNode) initializeSupervisionTree() (goakt.ActorSystem, error) {
...
 		return nil, err
 	}
-	_, err = actorSystem.Spawn(nn.ctx, actors.ControlAPIActorName, actors.CreateControlAPI(nn.nc, pk))
+
+	controlApiActor, err := actorSystem.Spawn(nn.ctx, actors.ControlAPIActorName,
+		actors.CreateControlAPI(nn.nc, nn.options.Logger, pk, nn.auctionResponse, nn.pingResponse, nn.infoResponse))
 	if err != nil {
 		return nil, err
 	}
+	nn.runningActors[actors.ControlAPIActorName] = controlApiActor
```
autodidaddict, line 246: remove please

autodidaddict (changes requested):
nits

node/node.go
jordan-rash: learning the hard way!

autodidaddict (approved):

PR #415, opened 2024-10-30
adds a tester agent

autodidaddict:


agents/tester/cmd/tester/main.go
```diff
@@ -0,0 +1,75 @@
...
+	return nil
+}
+
+func (a *testAgent) StopWorkload(req *agentapigen.StopWorkloadRequestJson) error {
+	delete(a.workloads, req.WorkloadId)
+	return nil
+}
+
+func (a *testAgent) ListWorkloads() error {
+	return nil
```
jordan-rash, line 74: maybe iterate through `a.workloads` and print with logger for now?

agents/common/nexagent.go
```diff
@@ -1,34 +1,169 @@
...
+
+	if err := agent.registerWithNode(); err != nil {
+		return err
+	}
+
+	if err := agent.createSubscriptions(); err != nil {
+		return err
+	}
+
+	switch os.Args[1] {
```
jordan-rash, line 69: What if we dont pass it a bunch of args here and instead we pass it a json string with all the parameters it's expecting?  Seems easier than parsing the command line

agents/tester/cmd/tester/main.go
autodidaddict: Nothing actually calls this function yet. When I do invoke it, that's probably what I'll have it do

agents/common/nexagent.go
autodidaddict: there's only 1 command line parameter here - `up` or `preflight`. The reason they're command line switches is so that they can be executed in a variety of ways, including by the `nex node preflight` command, which will eventually locate all the agents and run them

autodidaddict: every other parameter the agent needs comes from an environment variable

jordan-rash (approved):

PR #416, opened 2024-11-01
Direct Start Workload

jordan-rash:


node/actors/artifacts.go
```diff
@@ -24,24 +28,57 @@ type ArtifactReference struct {
...
 	}
 	return nil, nil
 }
 
 func cacheFile(name string, location *url.URL) (*ArtifactReference, error) {
-	// TODO: Implement
-	return nil, nil
+	filePath, tag := "", ""
+	if strings.Contains(location.Path, ":") {
+		sPath := strings.Split(location.Path, ":")
```
autodidaddict, line 58: for safety's sake, we should probably use `SplitN(s, ":", 1)`, which will avoid crashing things if the tag actually has a `:` in it (we should block that.. but safety iz gud)

node/actors/direct_start_agent.go
```diff
@@ -32,17 +44,99 @@ func (a *DirectStartAgent) PostStop(ctx context.Context) error {
 }
 
 func (a *DirectStartAgent) Receive(ctx *goakt.ReceiveContext) {
```
autodidaddict, line 46: the receive function is a bit busy... I think it's cleaner to stay high level in the receive function and then call out to helpers in the `switch`, making it easier to skim through the list of messages being handled.

node/actors/direct_start_agent.go
```diff
@@ -32,17 +44,99 @@ func (a *DirectStartAgent) PostStop(ctx context.Context) error {
...
+				}
+				ctx.Response(&actorproto.WorkloadStopped{Id: m.WorkloadId, Stopped: true})
+				return
+			}
+		}
+		a.logger.Error("workload not found", slog.String("name", ctx.Self().Name()), slog.String("workload", m.WorkloadId))
+		ctx.Err(errors.New("workload not found"))
 	case *actorproto.QueryWorkloads:
-		a.queryWorkloads(ctx)
+		ret := new(actorproto.WorkloadList)
```
autodidaddict, line 120: see previous comment... let's create a single helper function per `case` branch so the dispatch pattern is clear in the `Receive` handler.

node/actors/process_actor.go
```diff
@@ -0,0 +1,80 @@
...
+	return nil
+}
+
+func (a *processActor) Receive(ctx *goakt.ReceiveContext) {
+	switch ctx.Message().(type) {
+	case *goaktpb.PostStart:
+		a.self = ctx.Self()
+		ctx.Tell(ctx.Self(), &actorproto.SpawnDirectStartProcess{})
+	case *actorproto.SpawnDirectStartProcess:
+		go func() {
```
autodidaddict, line 54: let's move the function to a named function so it's a bit cleaner.

node/actors/process_actor.go
```diff
@@ -0,0 +1,80 @@
...
+				return
+			}
+		}()
+	case *actorproto.KillDirectStartProcess:
+		err := a.process.Stop("commanded")
+		if err != nil {
+			a.logger.Error("failed to stop process", slog.Any("err", err))
+			ctx.Err(err)
+			return
+		}
```
autodidaddict, line 68: here we should call `ctx.Shutdown()` to terminate the process actor cleanly (which should then also invoke code in the `PostStop` function during cleanup.

node/actors/process_actor.go
```diff
@@ -0,0 +1,80 @@
...
+	switch ctx.Message().(type) {
+	case *goaktpb.PostStart:
+		a.self = ctx.Self()
+		ctx.Tell(ctx.Self(), &actorproto.SpawnDirectStartProcess{})
+	case *actorproto.SpawnDirectStartProcess:
+		go func() {
+			err := a.process.Run()
+			if err != nil {
+				a.logger.Error("failed to start process", slog.Any("err", err))
+				ctx.Err(err)
```
autodidaddict, line 58: we might want to use `ctx.Shutdown` here, because if `run` didn't succeed, this actor needs to go away anyway.

autodidaddict (changes requested):

agents/common/nexagent.go
```diff
@@ -138,10 +138,12 @@ func (agent *NexAgent) handleStartWorkload(m *nats.Msg) {
 	var req agentapigen.StartWorkloadRequestJson
 	err := json.Unmarshal(m.Data, &req)
 	if err != nil {
+		_ = 0 // for linter
```
autodidaddict, line 141: an angel just lost its wings ;)

jordan-rash: its the best I could do 🤣

node/actors/direct_start_agent.go
```diff
@@ -32,17 +44,111 @@ func (a *DirectStartAgent) PostStop(ctx context.Context) error {
...
+		a.logger.Error("Failed to spawn child", slog.String("name", a.self.Name()), slog.String("workload", m.WorkloadName), slog.Any("err", err))
+		return nil, err
+	}
+
+	a.logger.Info("Spawned direct start process", slog.String("name", c.Name()))
+	return &actorproto.WorkloadStarted{Id: workloadId, Name: m.WorkloadName, Started: true}, nil
+}
+
+func (a *DirectStartAgent) stopWorkload(m *actorproto.StopWorkload) (*actorproto.WorkloadStopped, error) {
+	for _, c := range a.self.Children() {
```
autodidaddict, line 122: you don't need to range through all children here. Instead use `Child(workloadId)` to get the specific one that you want.

autodidaddict: did copilot give you this?

node/actors/os_procs.go
```diff
@@ -60,23 +61,35 @@ func (proc *OsProcess) Run() error {
...
 	}
+
+	proc.logger.Debug("OS process terminated successfully", slog.Int("pid", proc.cmd.Process.Pid))
+	proc.cmd = nil
+
+	return nil
+}
+
+// Sends an interrupt signal to the running process (if it's running)
+func (proc *OsProcess) Stop(reason string) error {
```
autodidaddict, line 72: Do we want to call this interrupt instead of `Stop` ? Is there a scenario where interrupt doesn't stop a process?

node/actors/os_procs.go
```diff
@@ -60,23 +61,35 @@ func (proc *OsProcess) Run() error {
...
-func (proc *OsProcess) Stop(reason string) error {
+func (proc *OsProcess) Kill() error {
 	if proc.cmd.Process != nil {
 		return proc.cmd.Process.Kill()
 	}
 	return nil
 }
 
+func (proc *OsProcess) IsRunning() bool {
+	return proc.cmd != nil
```
autodidaddict, line 90: might also want to add a check for proc.cmd.Process != nil

node/actors/process_actor.go
```diff
@@ -0,0 +1,115 @@
...
+	return nil
+}
+
+func (a *processActor) Receive(ctx *goakt.ReceiveContext) {
+	switch ctx.Message().(type) {
+	case *goaktpb.PostStart:
+		a.self = ctx.Self()
+		ctx.Tell(ctx.Self(), &actorproto.SpawnDirectStartProcess{})
+	case *actorproto.SpawnDirectStartProcess:
+		go a.Spawn(ctx)
```
autodidaddict, line 55: this is a confusing name. It looks like the goakt Spawn function... can we call it `Spawn???` where `???` is maybe something like `SpawnOsProcess` ?

node/actors/process_actor.go
```diff
@@ -0,0 +1,115 @@
...
+	case *actorproto.KillDirectStartProcess:
+		err := a.process.Stop("commanded")
```
autodidaddict, line 57: I still think there's too much code inside this case statement. Let's move the bulk of this to a helper function. As a design guide I like to have case branches of 5 lines or less. It's more tolerable in some other languages but it's super easy to lose your place in Go's switch statements.

autodidaddict (changes requested):
just a few minor things

agents/common/nexagent.go
autodidaddict: This is the linter's fault.

node/actors/os_procs.go
jordan-rash: So I found that even after the stop, the proc.cmd.Process was never nil...thats why I manually set `proc.cmd = nil` after cmd.Wait unblocks

autodidaddict: Huh. Learn something new every day

node/actors/artifacts.go
```diff
@@ -24,24 +28,57 @@ type ArtifactReference struct {
 	Size int
 }
 
+func extractTag(location *url.URL) string {
+	return location.Fragment
```
autodidaddict, line 32: What happened to the code that checks for a :, splits, and returns the tag and if none exists it returns "latest"?

node/actors/artifacts.go
```diff
@@ -24,24 +28,57 @@ type ArtifactReference struct {
...
+		return cacheOciArtifact(name, extractTag(location), location)
 	}
 	return nil, nil
 }
 
 func cacheFile(name string, location *url.URL) (*ArtifactReference, error) {
-	// TODO: Implement
-	return nil, nil
+	filePath, tag := "", ""
+	sPath := strings.SplitN(location.Path, ":", 2) // splits on first : only
```
autodidaddict, line 57: Should this call extract tag?

node/actors/process_actor.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		a.logger.Error("failed to stop process", slog.Any("err", err))
+		err := a.process.Kill()
+		if err != nil {
+			a.logger.Error("failed to kill process", slog.Any("err", err))
+			return err
+		}
+	}
+
+	// waits 5 seconds for workload to shutdown gracefully
```
autodidaddict, line 97: It pains me to see a timed retry loop appear here. We should eventually figure out a better way:tm:

autodidaddict (approved):

node/actors/artifacts.go
jordan-rash: Hasn't been written.  I'm going to do a PR for artifacts next. It'll address all the functionality

node/actors/artifacts.go
jordan-rash: Artifacts will have its own PR tomorrow

PR #418, opened 2024-11-04
file/oci/nats Artifact Support

jordan-rash:


jordan-rash:
This will fail until my access request is approved

autodidaddict:
it's hard to tell from the github diff...but this artifact caching code will be reusable when we go to grab the external agents (javascript, wasm, etc). They will have the same URL format in the node configuration file in the agents section. I just want to make sure we can reuse this code.

jordan-rash:
> it's hard to tell from the github diff...but this artifact caching code will be reusable when we go to grab the external agents (javascript, wasm, etc). They will have the same URL format in the node configuration file in the agents section. I just want to make sure we can reuse this code.

Correct.  There is nothing "direct start" about this.  As long as the uri is valid, any of the agents will be able to use the functionality

autodidaddict (approved):

PR #424, opened 2024-11-07
CLI Monitor Command

jordan-rash:
Ability to monitor workloads output 
```
└─❯ go run . monitor logs --workload-id rhvGqxdnXTLeXeZckA3WPJ
##### Subscribing to: $NEX.logs.rhvGqxdnXTLeXeZckA3WPJ.* | 07 Nov 24 10:18 MST
[0.2s] -> Count: 180
[1.2s] -> Count: 181
[2.2s] -> Count: 182
[3.2s] -> Count: 183
[4.2s] -> Count: 184
```


autodidaddict:
I think it might make sense to move the monitoring code into `api/nodecontrol` rather than hiding it behind the CLI package. This way, any client can be sure to monitor these things the right way, regardless of whether or not it's a CLI.

api/nodecontrol/client.go
```diff
@@ -186,3 +186,39 @@ func (c *ControlAPIClient) SetLameDuck(nodeId string) (*nodegen.LameduckResponse
 
 	return resp, nil
 }
+
+func (c *ControlAPIClient) MonitorLogs(workloadId, level string) (chan []byte, error) {
```
autodidaddict, line 190: since we know logs and events will have some standard / envelope structure, we should probably attempt to decode that inside the client so that the channels can be of type `*logentry` and `*cloudevent` (just examples).

autodidaddict (changes requested):

api/nodecontrol/client.go
```diff
@@ -186,3 +187,51 @@ func (c *ControlAPIClient) SetLameDuck(nodeId string) (*nodegen.LameduckResponse
...
+	if err != nil {
+		return nil, err
+	}
+
+	ret := make(chan *cloudevents.Event)
+	_, err = c.nc.Subscribe(f_subject, func(msg *nats.Msg) {
+		e := new(cloudevents.Event)
+		err := json.Unmarshal(msg.Data, e)
+		if err != nil {
+			c.logger.Error("failed to unmarshal log message", slog.Any("err", err), slog.String("data", string(msg.Data)))
```
autodidaddict, line 227: copy/paste issue. this is an event, not a log message.

node/internal/actors/os_procs.go
```diff
@@ -112,11 +114,16 @@ type logCapture struct {
 
 // Log capture implementation of io.Writer for stdout and stderr
 func (cap logCapture) Write(p []byte) (n int, err error) {
+	e := models.NewEnvelope[string](WorkloadLogType, string(p), 200, nil)
```
autodidaddict, line 117: I don't think we need an envelope here. We should just have a standard structure for log entries. The envelope from models has things like a result code which don't apply to logs.

api/nodecontrol/client.go
```diff
@@ -186,3 +187,45 @@ func (c *ControlAPIClient) SetLameDuck(nodeId string) (*nodegen.LameduckResponse
 
 	return &envelope.Data, nil
 }
+
+func (c *ControlAPIClient) MonitorLogs(workloadId, level string) (chan []byte, error) {
```
autodidaddict, line 191: At some point we're going to want to enable people to monitor events and logs from an entire namespace rather than a specific workload id.

cmd/nex/monitor.go
```diff
@@ -1,10 +1,122 @@
...
+	if e.EventType != "*" {
+		eWord = e.EventType
+	}
+	fmt.Printf("##### %s event types for application: %s\n", eWord, e.WorkloadID)
+
+logloop:
+	for {
+		select {
+		case <-ctx.Done():
+			break logloop
```
autodidaddict, line 65: Would a continue work here instead of a labeled break?

models/log_subjects.go
```diff
@@ -0,0 +1,47 @@
+package models
+
+import "fmt"
+
+const (
+	// $NEX.logs."WorkloadID"."stdout|stderr"
+	LOGS_SUBJECT NexSubject = "$NEX.logs.%s.%s"
+
+	// $NEX.events."WorkloadID"."EventType"
+	EVENTS_SUBJECT NexSubject = "$NEX.events.%s.%s"
```
autodidaddict, line 10: I didn't think Go preferred shouty constants versus just pascal case

autodidaddict (approved):

api/nodecontrol/client.go
jordan-rash: That is already in the works

cmd/nex/monitor.go
jordan-rash: it was originally like that, but I think it continued the select instead of breaking out of the for....I cant remember

PR #425, opened 2024-11-07
first use of agent registration

autodidaddict:
This doesn't actually include the code to retrieve and execute the external agent binary, but it gets close. It also illustrates the pattern for handling the registration messages that come from agents (which is very different from how `main` does agent handshakes).

Also shows the use of `Become` to change the actor behavior from pre-registration to post-registration, which keeps the number of superfluous `if` statements down.


node/internal/actors/external_agent.go
```diff
@@ -34,16 +52,77 @@ func (a *ExternalAgent) PostStop(ctx context.Context) error {
...
+}
+
+func (a *ExternalAgent) createNatsConnection(url string, _ string, seed string) {
+	var err error
+	opt, err := nats.NkeyOptionFromSeed(seed)
+	if err != nil {
+		a.logger.Error("Failed to extract an nkey option from the nkey seed", slog.Any("error", err))
+		return
+	}
+	a.conn, err = nats.Connect(url, opt)
```
jordan-rash, line 117: can we add the "name" option here so the connections are distinguished more easily

autodidaddict: Sure but it's a connection to a server no one knows about nor can they connect to it. 

Will add the name tomorrow

PR #426, opened 2024-11-07
Lameduck Mode

jordan-rash:
From CLI 
```
└─❯ go run . node lameduck --node-id NAU55EVRD3Y3ACS5PUQWJEJFAWREEZPADZEFVYQGIBR7IT3I5QJFYSIE
Node NAU55EVRD3Y3ACS5PUQWJEJFAWREEZPADZEFVYQGIBR7IT3I5QJFYSIE is now in lameduck mode. Workloads will begin stopping gracefully.
```

Server logs
```
<..SNIP..>
[INF] 2024-11-07 14:03:15 - NATS execution engine awaiting commands id=NAU55EVRD3Y3ACS5PUQWJEJFAWREEZPADZEFVYQGIBR7IT3I5QJFYSIE
[INF] 2024-11-07 14:03:15 - Control API NATS server is running name=control_api
[INF] 2024-11-07 14:03:38 - Received lame duck request
[INF] 2024-11-07 14:03:39 - Shutting down nexnode
agent_supervisor | [DBG] 2024-11-07 14:03:39 - Agent supervisor stopped
```


node/internal/actors/control_api.go
```diff
@@ -45,11 +47,12 @@ type ControlAPI struct {
 	auctionResponse  auctionResponseFunc
 	nodePingResponse nodePingResponseFunc
 	nodeInfoResponse nodeInfoResponseFunc
+	nodeLameDuck     nodeLameDuckFunc
 
 	self *goakt.PID
 }
 
-func CreateControlAPI(nc *nats.Conn, logger *slog.Logger, publicKey string, aF auctionResponseFunc, nF nodePingResponseFunc, iF nodeInfoResponseFunc) *ControlAPI {
+func CreateControlAPI(nc *nats.Conn, logger *slog.Logger, publicKey string, aF auctionResponseFunc, nF nodePingResponseFunc, iF nodeInfoResponseFunc, ldF nodeLameDuckFunc) *ControlAPI {
```
autodidaddict, line 55: rather than passing in 4 different functions, we should probably extract that into a callback interface and just pass the one thing. This is the pattern being used in the agent common interface.

autodidaddict (changes requested):

node/internal/actors/control_api.go
jordan-rash: okay....this slowly grew from 1 to 2 to 3 to 4 functions over a week...so good point.  Will look into interface

api/nodecontrol/lameduck-response.json
```diff
@@ -4,15 +4,15 @@
   "title": "LameduckResponse",
   "type": "object",
   "properties": {
-    "node_id": {
+    "id": {
       "type": "string",
-      "description": "The unique identifier of the node"
+      "description": "The unique identifier of the actor"
```
autodidaddict, line 9: the outside world should never know about the existence of actors. This ID probably doesn't need to be here at all, since the lame duck request would've been sent to a specific node.

node/internal/actors/agent_supervisor.go
```diff
@@ -66,3 +72,23 @@ func (s *AgentSupervisor) queryWorkloads(ctx *goakt.ReceiveContext) {
...
+			return &actorproto.LameDuckResponse{Id: ctx.Self().ID(), Success: false}
+		}
+		resp, ok := r.(*actorproto.LameDuckResponse)
+		if !ok {
+			s.logger.Warn("Unexpected response from agent")
+			return &actorproto.LameDuckResponse{Id: ctx.Self().ID(), Success: false}
+		}
+		if !resp.Success {
+			s.logger.Error("Child workload failed to shutdown", slog.String("child", pid.ID()), slog.Any("err", err))
+			return &actorproto.LameDuckResponse{Id: ctx.Self().ID(), Success: false}
```
autodidaddict, line 90: absolutely never expose an actor's ID anywhere outside the internal actor plumbing.

autodidaddict (changes requested):
need to not expose an actor ID to a public api

autodidaddict (approved):

PR #433, opened 2024-11-15
Attempting to make direct start os/arch aware

jordan-rash:
```
go run . workloads build --binaries /tmp/counter/counter,/tmp/counter/osx/counter --push --registry ghcr.io --name jordan-rash/derp --registry-user jordan-rash --registry-password $GHCR_TOKEN --tag multi
```

`workload bundle` attempts to detect what os/arch a binary is and tags the artifact layer appropriately.  Then, when an OCI artifact is being pulled, it will only start if GOOS/GOARCH pair match a layer in artifact


cmd/nex/workload.go
```diff
@@ -206,10 +209,10 @@ func (s StopWorkload) Run(ctx context.Context, globals *Globals, w *Workload) er
 }
 
 type BundleWorkload struct {
-	Binary string `description:"Binary to package"`
-	OS     string `description:"Operating system of the binary" enum:"linux,darwin" default:"linux"`
-	Arch   string `description:"Architecture of the binary" enum:"amd64,arm64" default:"amd64"`
-	Output string `description:"Output file name" default:"./artifact.tar"`
+	Binaries []string `description:"Binary to package"`
```
autodidaddict, line 212: does this mean that I can bundle multiple x86 binaries and multiple x64 binaries with 2 consecutive bundle commands? To build an image with 1 artifact per OS/arch combo, would I make `n` bundle commands?

autodidaddict (approved):

cmd/nex/workload.go
jordan-rash: you can add as many binaries as you want to the bundle.  So there could technically be 1000 linux/amd64 artifacts in there.  At this time though, `direct_start` will start the first one it finds

PR #434, opened 2024-11-18
WIP: Auction on Deploy

jordan-rash:
This work allows a deploy request to be sent without specifying a node.  In order to accomplish this, an auction is done and a node it found in the ClientAPI based on tags.  If multiple nodes are returned in action, one is picked at random. 

Since the user does not know what node holds their workload, in order to stop, a workload ping will find the workload during a stop request.  This work is not completed yet.

```
└─❯ go run . workload run --name counter --uri oci://localhost:5000/demo:latest
Workload counter [w49rOf145nNI6HAlvg8IJB] started on node
```


node/node.go
```diff
@@ -421,3 +421,14 @@ func (nn nexNode) SetLameDuck(ctx context.Context) {
 		nn.interrupt <- os.Interrupt
 	}()
 }
+
+func (nn nexNode) IsTargetNode(inId string) (bool, error) {
+	pub, err := nn.publicKey.PublicKey()
+	if err != nil {
+		return false, err
+	}
+	if inId == pub || inId == nn.auctionBidId {
```
autodidaddict, line 430: Using a single auction bid ID here is problematic. For example, if two different clients perform an auction, then when the second one issues a deploy request, the bid ID will be gone because the node has only a single bid ID.

In short, _every_ reply to an auction should be a unique bid ID. The bid IDs can be stored in a map between the auction request and the positive bid. Maybe this means each auction needs its own nuid as well, so we can create a `map[bidId]auctionID`.

The node can expire these bid IDs every 30s or so. When a deploy/run request comes into the node for a given auction ID, if there's no corresponding auction ID, then the deploy request is rejected.

jordan-rash: I literally realized this as I was getting in bed last night 🤣

api/nodecontrol/client.go
```diff
@@ -118,6 +134,40 @@ func (c *ControlAPIClient) FindWorkload(_type, namespace, workloadId string) (*n
...
+	nodeIdx := rand.IntN(len(auctionResults))
+	// This "NodeID" is acutally the BidderID
+	node := auctionResults[nodeIdx].NodeId
+
+	req_b, err := json.Marshal(req)
+	if err != nil {
+		return nil, err
+	}
+
+	msg, err := c.nc.Request(models.AuctionDeployRequestSubject(namespace, node), req_b, DefaultRequestTimeout)
```
autodidaddict, line 157: this shouldn't be `node` here. This is the auction ID that we chose, _not_ the node ID.

node/internal/actors/control_api.go
```diff
@@ -212,6 +220,55 @@ func (api *ControlAPI) handleAuction(m *nats.Msg) {
...
+
+	req := new(nodecontrol.StartWorkloadRequestJson)
+	err := json.Unmarshal(m.Data, req)
+	if err != nil {
+		api.logger.Error("Failed to unmarshal deploy request", slog.Any("error", err))
+		models.RespondEnvelope(m, RunResponseType, 500, "", fmt.Sprintf("failed to unmarshal deploy request: %s", err))
+		return
+	}
+
+	target, err := api.nodeCallback.IsTargetNode(splitSub[3])
```
autodidaddict, line 235: remember that this needs to be the _auctionId_ and not a node ID.

autodidaddict (changes requested):
we need to not use the nodeID in the last token of the `ADEPLOY` subject.

api/nodecontrol/auction-response.json
```diff
@@ -4,7 +4,7 @@
   "title": "AuctionResponse",
   "type": "object",
   "properties": {
-    "node_id": {
+    "bidder_id": {
       "type": "string",
       "description": "The unique identifier of the node"
```
autodidaddict, line 9: nit.. this could be something like `A one-time identifier used to target deployments`

jordan-rash: 🤣 Translation: _When you change the field, update the description too dummy_

node/node.go
```diff
@@ -428,7 +429,12 @@ func (nn nexNode) IsTargetNode(inId string) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	if inId == pub || nn.auctionMap.Exists(inId) {
+	if inId == pub {
+		return true, nil
+	}
+	if nn.auctionMap.Exists(inId) {
+		auctionId := nn.auctionMap.Get(inId)
+		nn.options.Logger.Debug("Accepting auction", slog.String("auctionId", auctionId))
```
autodidaddict, line 437: nit: should be something like "accepting workload from auction" , otherwise this looks like the first step in an auction not the 2nd

autodidaddict (dismissed):

PR #435, opened 2024-11-19
Lameduck delay

jordan-rash:
This adds a delay to lame duck that waits to start shutting down workloads. The node will go into "Lameduck Mode" immediately, just the shutdown of workloads is delayed. 

```
└─❯ go run . node lameduck --node-id NAU55EVRD3Y3ACS5PUQWJEJFAWREEZPADZEFVYQGIBR7IT3I5QJFYSIE --delay 10s
Node NAU55EVRD3Y3ACS5PUQWJEJFAWREEZPADZEFVYQGIBR7IT3I5QJFYSIE is now in lameduck mode. Workloads will begin stopping gracefully in 10s.

{FROM SERVER LOGS}
[WRN] 2024-11-19 15:14:16 - Lameduck mode enabled. Will start shutting down workloads. workload_shutdown=2024-11-19 15:14:26

{NOTE * NEXT TO NODEID}
└─❯ go run . node ls
╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                               NATS Execution Nodes                                                               │
├───────┬───────────────────────────────────────────────────────────┬───────────────┬─────────┬────────┬───────────┬───────────────┬───────┬───────┤
│ Nexus │ ID (* = Lameduck Mode)                                    │ Name          │ Version │ Agents │ Workloads │ Uptime        │ OS    │ Arch  │
├───────┼───────────────────────────────────────────────────────────┼───────────────┼─────────┼────────┼───────────┼───────────────┼───────┼───────┤
│ nexus │ NAU55EVRD3Y3ACS5PUQWJEJFAWREEZPADZEFVYQGIBR7IT3I5QJFYSIE* │ primary-suite │ 0.0.0   │      1 │ 0         │ 14.283339635s │ linux │ amd64 │
╰───────┴───────────────────────────────────────────────────────────┴───────────────┴─────────┴────────┴───────────┴───────────────┴───────┴───────╯
```


autodidaddict (approved):
As long as the node goes into lame duck mode immediately and the delay is between that and the removal of the first workload, I like this. We just don't want to delay switching into lame duck mode because we'll need that buffer time to reject incoming start requests.

jordan-rash:
> As long as the node goes into lame duck mode immediately and the delay is between that and the removal of the first workload, I like this. We just don't want to delay switching into lame duck mode because we'll need that buffer time to reject incoming start requests.

💯

PR #438, opened 2024-11-20
API Remodel

jordan-rash:
Makes it so that the API is more scopeable(TM) by namespace


api/nodecontrol/client.go
```diff
@@ -122,7 +107,7 @@ func (c *ControlAPIClient) DirectPing(nodeId string) (*nodegen.NodePingResponseJ
 }
 
 func (c *ControlAPIClient) FindWorkload(_type, namespace, workloadId string) (*nodegen.WorkloadPingResponseJson, error) {
-	msg, err := c.nc.Request(models.AgentPingWorkloadRequestSubject(_type, namespace, workloadId), nil, DefaultRequestTimeout)
+	msg, err := c.nc.Request(models.WorkloadPingRequestSubject(_type, namespace, workloadId), nil, DefaultRequestTimeout)
```
autodidaddict, line 110: If we're actually using this variable, we should remove the `_` ... Weird that the linter doesn't complain about this when it's so cranky about the inverse.

api/nodecontrol/client.go
```diff
@@ -175,7 +159,7 @@ func (c *ControlAPIClient) DeployWorkload(namespace, nodeId string, req nodegen.
 		return nil, err
 	}
 
-	msg, err := c.nc.Request(models.DeployRequestSubject(namespace, nodeId), req_b, DefaultRequestTimeout)
+	msg, err := c.nc.Request(models.DeploySubject(nodeId), req_b, DefaultRequestTimeout)
```
autodidaddict, line 162: should we call this a "direct deploy" to contrast against the "auction deploy"

autodidaddict: could even change the subject from `DEPLOY` to `DDEPLOY` to contrast `ADEPLOY`

jordan-rash: I can do that...

api/nodecontrol/client.go
jordan-rash: its because `type` is a reserved word.....I can call it something else

autodidaddict: oh, hah.. makes sens. yeah, maybe `workloadType`

autodidaddict (approved):

PR #440, opened 2024-11-22
javascript runner

autodidaddict:
This sets up the ability to run javascript functions (must be named `run`) lambda-style with a host services provider. It does not yet complete the circle of life so scripts won't yet be executable front to back, but I think that will get buttoned up in the next PR. Wanted to keep this one focused on just the script execution.

### DIFF
--- a/agents/javascript/script_runner_test.go
+++ b/agents/javascript/script_runner_test.go
@@ -1,0 +1,142 @@
+package agent
+
+import (
+	"errors"
+	"fmt"
+	"testing"
+
+	"github.com/dop251/goja"
+	"github.com/nats-io/nuid"
+)
+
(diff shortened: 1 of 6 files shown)


agents/javascript/script_runner_test.go
```diff
@@ -0,0 +1,142 @@
...
+	"errors"
+	"fmt"
+	"testing"
+
+	"github.com/dop251/goja"
+	"github.com/nats-io/nuid"
+)
+
+func setupSuite(t testing.TB) (*ScriptRunner, string, func(tb testing.TB)) {
+	// I love how this would make the Rust compiler send a hit squad after me
```
jordan-rash, line 13: I think you need t.Helper() at the top of this

PR #469, opened 2024-12-26
Make `system` namespace a const

### LINKED ISSUE #468: Make `system` namespace a const
`system` represents a special namespace that has additional administrative capabilities.  There are multiple places in the codebase is it manually typed out as opposed to using a const.  This needs to be updated to use a const
           
> I think I have seen a number of places where `"system"` is referred. Suggest making a standard constant and referencing that, e.g. `const systemNamespace = "system"` or something.

_Originally posted by @bruth in https://github.com/synadia-io/nex/pull/461#discussion_r1895705464_

jordan-rash:
Closes #468

### DIFF
--- a/models/node.go
+++ b/models/node.go
@@ -1,5 +1,10 @@
 package models
 
 const (
+	// This is a special namespace that is used for "admin" only requests
+	// The node listens on a handful of topics for and typically the
+	// 3rd position is the namespace.  There are some endpoints that can be seen in
+	// `node/internal/actors/subjects.go` that have "system" hardcoded in
+	// the third position.  These are considered privledged endpoints
 	NodeSystemNamespace string = "system"
 )
--- a/models/node.go
+++ b/models/node.go
@@ -1,0 +1,5 @@
+package models
+
+const (
+	NodeSystemNamespace string = "system"
+)
--- a/node/node.go
+++ b/node/node.go
@@ -205,7 +205,7 @@ func (nn *nexNode) initializeSupervisionTree() error {
 	var err error
 
 	nn.actorSystem, err = goakt.NewActorSystem("nexnode",
-		goakt.WithLogger(logger.NewSlog(nn.options.Logger.Handler().WithGroup("system"))),
+		goakt.WithLogger(logger.NewSlog(nn.options.Logger.Handler().WithGroup("actor_system"))),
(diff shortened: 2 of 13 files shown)


node/node.go
```diff
@@ -205,7 +205,7 @@ func (nn *nexNode) initializeSupervisionTree() error {
 	var err error
 
 	nn.actorSystem, err = goakt.NewActorSystem("nexnode",
-		goakt.WithLogger(logger.NewSlog(nn.options.Logger.Handler().WithGroup("system"))),
+		goakt.WithLogger(logger.NewSlog(nn.options.Logger.Handler().WithGroup("actor_system"))),
```
bruth, line 208: Curious why the addition of the `actor_` prefix?

models/node.go
```diff
@@ -0,0 +1,5 @@
+package models
+
+const (
+	NodeSystemNamespace string = "system"
```
bruth, line 4: Worth adding a comment to indicate some more context on what the "system" means in this context. I saw above you added a template variable called `adminNamespace`.

bruth (commented):
One comment and one question.

node/node.go
jordan-rash: This is to hide the internal goakt logs that we dont want.  There is no other way to turn them off, so I just filter them out and provide a `--show-system-logs` flag to view them if you want

models/node.go
jordan-rash: yea, I named it admin in the CLI because I though that was more clear, but I didn't want to name it "admin" in the internals.  I will add some comments

node/node.go
jordan-rash: Forgot to answer your question....

I added the `actor_` prefix because this PR started using the word "system" and something real and I didn't want the two to be confused

jordan-rash: [2ab1cef](https://github.com/synadia-io/nex/pull/469/commits/2ab1cefc4bfbf22af78cf44813405821418d4101)

models/node.go
jordan-rash: [2ab1cef](https://github.com/synadia-io/nex/pull/469/commits/2ab1cefc4bfbf22af78cf44813405821418d4101)

bruth (approved):

PR #473, opened 2024-12-27
Missing Control API Unit Tests

jordan-rash:

### DIFF
--- a/api/nodecontrol/client_test.go
+++ b/api/nodecontrol/client_test.go
@@ -841,3 +841,103 @@ func TestCopyWorkload(t *testing.T) {
 	}
 	t.Log(*cResp)
 }
+
+func TestMonitorEndpoints(t *testing.T) {
+	workingDir := t.TempDir()
+	natsServer, err := startNatsServer(t, workingDir)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+
+	t.Cleanup(func() {
+		os.RemoveAll(filepath.Join(os.TempDir(), "inex-NCUU2YIYXEPGTCDXDKQR7LL5PXDHIDG7SDFLWKE3WY63ZGCZL2HKIAJT"))
+		cancel()
+		natsServer.Shutdown()
+	})
+
+	stdout := new(bytes.Buffer)
+	stderr := new(bytes.Buffer)
+	logger := slog.New(shandler.NewHandler(
+		shandler.WithLogLevel(slog.LevelDebug),
+		shandler.WithGroupFilter([]string{"actor_system"}),
+		shandler.WithStdOut(stdout),
+		shandler.WithStdErr(stderr),
+	))
+
+	err = startNexus(t, ctx, logger, workingDir, natsServer.ClientURL(), 1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	time.Sleep(1000 * time.Millisecond)
+	nc, err := nats.Connect(natsServer.ClientURL())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	control, err := NewControlApiClient(nc, logger)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	logs, err := control.MonitorLogs("system", "*", "*")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	logBuf := new(bytes.Buffer)
+	go func() {
+		for m := range logs {
+			_, err := logBuf.Write(m)
+			if err != nil {
+				t.Error(err)
+			}
+		}
+	}()
+
+	time.Sleep(250 * time.Millisecond)
+	err = nc.Publish("$NEX.logs.system.b.c", []byte("log test"))
+	if err != nil {
+		t.Fatal(err)
+	}
+	time.Sleep(250 * time.Millisecond)
+	close(logs)
+
+	// ----
+	events, err := control.MonitorEvents("system", "*", "*")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	eventBuf := new(bytes.Buffer)
+	go func() {
+		for m := range events {
+			mB, err := json.Marshal(m)
+			if err != nil {
+				t.Error(err)
+			}
+			_, err = eventBuf.Write(mB)
+			if err != nil {
+				t.Error(err)
+			}
+		}
+	}()
+
+	time.Sleep(250 * time.Millisecond)
+	err = nc.Publish("$NEX.events.system.b.c", []byte("{\"test\": \"event\"}"))
+	if err != nil {
+		t.Fatal(err)
+	}
+	time.Sleep(250 * time.Millisecond)
+	close(events)
+
+	if logBuf.String() != "log test" {
+		t.Fatalf("expected log test, got %s", string(logBuf.String()))
+	}
+
+	if eventBuf.String() != "{\"test\":\"event\"}" {
+		t.Fatalf("expected event {\"test\":\"event\"}, got %s", eventBuf.String())
+	}
+}
--- a/api/nodecontrol/client_test.go
+++ b/api/nodecontrol/client_test.go
@@ -5,20 +5,22 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"log/slog"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"testing"
 	"time"
 
 	"disorder.dev/shandler"
+	"github.com/carlmjohnson/be"
 	"github.com/nats-io/nats-server/v2/server"
 	"github.com/nats-io/nats.go"
 	"github.com/nats-io/nkeys"
+
 	"github.com/synadia-io/nex/api/nodecontrol/gen"
 	"github.com/synadia-io/nex/models"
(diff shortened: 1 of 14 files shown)


api/nodecontrol/client_test.go
```diff
@@ -0,0 +1,943 @@
...
+	workingDir := t.TempDir()
+	natsServer, err := startNatsServer(t, workingDir)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+
+	t.Cleanup(func() {
+		os.RemoveAll(filepath.Join(os.TempDir(), "inex-NCUU2YIYXEPGTCDXDKQR7LL5PXDHIDG7SDFLWKE3WY63ZGCZL2HKIAJT"))
```
bruth, line 147: Since the working directory is created from `t.TempDir()`, it will be cleaned up automatically when the test ends. No need for a separate cleanup call.

api/nodecontrol/client_test.go
```diff
@@ -0,0 +1,943 @@
...
+		shandler.WithStdOut(stdout),
+		shandler.WithStdErr(stderr),
+	))
+
+	err = startNexus(t, ctx, logger, workingDir, natsServer.ClientURL(), 1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	time.Sleep(1000 * time.Millisecond)
```
bruth, line 166: I presume this is waiting for the Nexus to start up? I would suggest using a wait group or channel in `startNexus` that blocks until the nodes are up (until some max time) or fail within the function.

api/nodecontrol/client_test.go
```diff
@@ -0,0 +1,943 @@
...
+	"github.com/nats-io/nkeys"
+	"github.com/synadia-io/nex/api/nodecontrol/gen"
+	"github.com/synadia-io/nex/models"
+	"github.com/synadia-io/nex/node"
+)
+
+const (
+	// NCUU2YIYXEPGTCDXDKQR7LL5PXDHIDG7SDFLWKE3WY63ZGCZL2HKIAJT - pub key
+	Node1ServerSeed string = "SNAB2T3VG2363NDA2JK7NT5O3FN5VCXI2MYJHOPFO2NIDXQU6DIWQTBQC4"
+	// XAL54S5FE6SRPONXRNVE4ZDAOHOT44GFIY2ZW33DHLR2U3H2HJSXXRKY - pub xkey
```
bruth, line 30: Since these are used below, I suggest making them constants that can be used rather than using the strings.

api/nodecontrol/client_test.go
```diff
@@ -0,0 +1,943 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	control, err := NewControlApiClient(nc, logger)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	resp, err := control.Auction("system", map[string]string{})
```
bruth, line 177: There is a new system namespace constant that could be used here correct?

api/nodecontrol/client_test.go
```diff
@@ -0,0 +1,943 @@
+package nodecontrol
```
bruth, line 1: I suggest adding a file or internal package that defines a set of standard assertion functions like [nats-server](https://github.com/nats-io/nats-server/blob/62de2cd3287f0ae21255f08c92efe379b318fa42/server/test_test.go#L58-L169). It shortens the tests a bit and keeps the output messages for simple asserts consistent.

api/nodecontrol/client_test.go
```diff
@@ -0,0 +1,943 @@
...
+)
+
+const (
+	// NCUU2YIYXEPGTCDXDKQR7LL5PXDHIDG7SDFLWKE3WY63ZGCZL2HKIAJT - pub key
+	Node1ServerSeed string = "SNAB2T3VG2363NDA2JK7NT5O3FN5VCXI2MYJHOPFO2NIDXQU6DIWQTBQC4"
+	// XAL54S5FE6SRPONXRNVE4ZDAOHOT44GFIY2ZW33DHLR2U3H2HJSXXRKY - pub xkey
+	Node1XKeySeed string = "SXAOUP7RZFW5QPE2GDWTPABUDM5UIAK6BPULJPWZQAFFL2RZ5K3UYWHYY4"
+)
+
+func buildTestBinary(t testing.TB, binMain string, workingDir string) (string, error) {
```
bruth, line 34: For test helper functions, fatal errors should not be returned and instead handled in the function. For example, if the `go build` command fails below, you can call `t.Fatal` directly (or if you introduce a `test.NoError` helper function). This avoids for the calling function (usually the actual test) to have to handle it.

models/node_options.go
```diff
@@ -40,6 +31,14 @@ type NodeOptions struct {
 
 type NodeOption func(*NodeOptions)
 
+func WithContext(ctx context.Context) NodeOption {
+	return func(n *NodeOptions) {
+		if ctx != nil {
```
bruth, line 36: Since `n.Context` is nil by default, is this check here to prevent overriding an existing non-nil context?

api/nodecontrol/client_test.go
```diff
@@ -0,0 +1,943 @@
...
+	resp, err := control.Auction("system", map[string]string{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(resp) != 1 {
+		t.Fatalf("expected 1 response, got %d", len(resp))
+	}
+
+	resp, err = control.Auction("system", map[string]string{"nex.node": "notreal"})
```
bruth, line 186: Can `TagNodeName` be used here?

api/nodecontrol/client_test.go
```diff
@@ -0,0 +1,943 @@
...
+
+	resp, err := control.GetInfo("NCUU2YIYXEPGTCDXDKQR7LL5PXDHIDG7SDFLWKE3WY63ZGCZL2HKIAJT", gen.NodeInfoRequestJson{
+		Namespace: "system",
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if resp.NodeId != "NCUU2YIYXEPGTCDXDKQR7LL5PXDHIDG7SDFLWKE3WY63ZGCZL2HKIAJT" {
+		t.Fatalf("expected node id NCUU2YIYXEPGTCDXDKQR7LL5PXDHIDG7SDFLWKE3WY63ZGCZL2HKIAJT, got %s", resp.Tags.Tags["nex.node"])
```
bruth, line 658: Should this be `, got %s", resp.NodeId`?

bruth (changes requested):
The bulk of the review are comments on a handful of practices to improve the testing structure which will help simplify the tests and remove some redundancy.

api/nodecontrol/client_test.go
jordan-rash: the `inex-*` directory is actually in os.TempDir and not configurable at this point...it is designed to persist incase of a node failure

bruth: To confirm this doesn't correspond to the `WithResourceDirectory` option for the node?

jordan-rash: It does not....it probably should

jordan-rash: https://github.com/synadia-io/nex/issues/474

api/nodecontrol/client_test.go
jordan-rash: Yea, when I was writing this, I realized I need a "healthy" check....this will probably be a function of the events subsystem that doesnt exist yet 😞

jordan-rash: Issue for tracking: https://github.com/synadia-io/nex/issues/475

models/node_options.go
jordan-rash: nope, thats a copy/paste error from the logger option right below it....removing

api/nodecontrol/client_test.go
jordan-rash: [126eead](https://github.com/synadia-io/nex/pull/473/commits/126eeadbf2414db6e2baa1c6eabd7f6c46bb250c) brings in [github.com/carlmjohnson/be](http://github.com/carlmjohnson/be)

api/nodecontrol/client_test.go
jordan-rash: [126eead](https://github.com/synadia-io/nex/pull/473/commits/126eeadbf2414db6e2baa1c6eabd7f6c46bb250c) brings in [github.com/carlmjohnson/be](http://github.com/carlmjohnson/be)

bruth (approved):
LGTM. My only other comment would be whether it would make sense to consolidate some of the tests as subtests to remove the need to spin up all the boilerplate each time for the simpler interactions. For example, `Ping` and `DirectPing` are not state changing, so those could combined as subtests. As another example, the only difference between the `AuctionDeployAnd*` tests is finding the workload vs. listing, but all the other boilerplate is the same. The `CopyWorkload` test could also be a subtest.

jordan-rash:
> LGTM. My only other comment would be whether it would make sense to consolidate some of the tests as subtests to remove the need to spin up all the boilerplate each time for the simpler interactions. For example, `Ping` and `DirectPing` are not state changing, so those could combined as subtests. As another example, the only difference between the `AuctionDeployAnd*` tests is finding the workload vs. listing, but all the other boilerplate is the same. The `CopyWorkload` test could also be a subtest.

Now that there is an assertion lib in there, I agree with this.  I thought before subtest made it too difficult to read....let me see if I can clean it up

PR #477, opened 2025-01-02
Migrate to assertion package

jordan-rash:
This PR moves existing tests to use the assertion package.
Additionally, it moves a handful of test helper functions to the `test` package

### DIFF
--- a/api/nodecontrol/client_test.go
+++ b/api/nodecontrol/client_test.go
@@ -5,123 +5,31 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/json"
-	"fmt"
 	"log/slog"
 	"os"
-	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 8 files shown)


bruth:
It looks like there are some tests in the source packages (e.g. `api/nodecontrol`) and some in `test`. Generally, it is more conventional to co-locate in the same package for where the source exists. There can be exceptions in the case of circular imports (which is a problem itself) or to prevent use of non-exported elements in the tests. Was there a specific motivation for splitting them up the way they are?

I like that the helpers have been extracted out into a package.

jordan-rash:
the `test/` package is not using any functions, I build all the binaries and interact as a user....thats all that is suppose to be in there.  The unit tests that call the functions directly live next to the code.

bruth (approved):

PR #519, opened 2026-05-19
adds reserved prefix for metadata

jordan-rash:


models/node.go
```diff
@@ -16,6 +16,8 @@ const (
 
 	AgentEnvNatsUrl = "NEX_AGENT_NATS_URL"
 	AgentEnvNodeId  = "NEX_AGENT_NODE_ID"
+
+	ReservedMetadataPrefix = "nex_"
```
joeriddles, line 20: What about using something like `$NEX.` ? inline with nats subjects (though this isn't a subject)

jordan-rash: Id prefer to keep map keys in this format. the subject style formatting feels wrong here.

PR #521, opened 2026-07-02
fix: keep NATS connections alive across credential expiry/rotation

Jarema:
## Problem
Node and agent NATS connections authenticate with a one-shot JWT+seed and set `MaxReconnects(-1)` but not `IgnoreAuthErrorAbort()`. When the credential expires or rotates, nats.go's `processAuthError` aborts the reconnect loop after two identical auth errors → the connection is permanently CLOSED and the node/agent is dead until manual restart (observed in production).

## Fix (4 commits)
- `232a33d` — extract `natsConnectionOptions()` (node.go + sdk/go/agent/nats.go), add `nats.IgnoreAuthErrorAbort()` + disconnect/reconnect/closed/error slog handlers. Keeps connections reconnecting through auth errors; initial-connect semantics unchanged.
- `7dfd6e7` — reloadable creds seam: when `NEX_NODE_NATS_CREDS_FILE` / `NEX_AGENT_NATS_CREDS_FILE` is set, authenticate via `nats.UserCredentials(path)` (re-read on every reconnect). Static JWT/seed fallback unchanged when unset. The producer that re-mints the creds file is deliberately deferred (`TODO(cred-refresh)` in node.go).
- `febcb47` — `nats.NoCallbacksAfterClientClose()` so intentional `Close()/Drain()` doesn't emit misleading disconnect/closed warnings.
- `31d9179` — e2e tests: deterministic TCP-mock proof (pre-fix options CLOSE after 2 auth errors; production options keep reconnecting) + live operator-mode nats-server test (short-TTL creds → expiry disconnect → rotate file → connection self-heals and round-trips; without rotation it keeps retrying and never goes CLOSED).

## Verification
`go build`, `go test -race ./...` (incl. new e2e, `-count=2`), `go vet`, `gofmt` all clean. Adversarial multi-agent review passed; all confirmed findings addressed in this stack.

Backport for the nex-ce 0.1.8 line (node half, sdk/ didn't exist at that pin): branch `hotfix-010-0.1.8`, fully green at `92b0e08`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/creds_rotation_e2e_test.go
+++ b/creds_rotation_e2e_test.go
@@ -1,0 +1,440 @@
+package nex
+
+import (
+	"bufio"
+	"fmt"
+	"net"
+	"os"
+	"path/filepath"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/carlmjohnson/be"
+	"github.com/nats-io/jwt/v2"
+	"github.com/nats-io/nats-server/v2/server"
+	"github.com/nats-io/nats.go"
+	"github.com/nats-io/nkeys"
+
+	"github.com/synadia-io/nex/internal/credentials"
+	"github.com/synadia-io/nex/models"
+)
+
+// ---------------------------------------------------------------------------
+// LAYER 1: deterministic auth-error reconnect behavior.
+//
+// A raw TCP mock server speaks just enough of the NATS protocol to (a) let the
+// initial Connect() succeed, then (b) return a repeated auth error on every
+// reconnect attempt. This isolates the reconnect/abort logic from JWT timing so
+// the test is fast and deterministic (blueprint: nats.go TestExpiredAuthentication).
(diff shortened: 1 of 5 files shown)


Jarema:
Note on the transient `test (ubuntu-latest)` failure on the first run: it was a data race between `(*NexNode).handlePing` (handlers.go:68) and `(*NexNode).Shutdown` (node.go:373) in the client lameduck tests — both untouched by this PR (diff is `natsConnectionOptions` + tests only). The base commit passes the same suite locally with `-race -count=3`, and the CI rerun is green. Looks like a pre-existing flake worth its own issue.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

joeriddles (approved):

piotrpio (approved):
LGTM

PR #522, opened 2026-07-02
fix: retry credential acquisition with exponential backoff

Jarema:
## Problem

Every credential-acquisition path in nex was one-shot — zero retries, zero backoff:

| Path | Location | On failure |
|------|----------|-----------|
| Node startup mint for embedded/local agents | `node.go` | log + skip — agent silently never starts |
| Handler mints (start-workload, agent registration, resume state, remote registration) | `handlers.go` (4 sites) | request fails; resume-state workloads silently dropped |
| Agent SDK `RemoteAgentInit` | `sdk/go/agent/runner.go` | single 3s request, error bubbles up |
| Agent SDK `Run` registration (where the nexlet receives its creds) | `sdk/go/agent/runner.go` | single request → process exits |
| AgentWatcher restarts (the only existing "retry") | `internal/watcher.go` | up to 3 restarts with **zero delay** — all burn inside the same transient blip, then gives up permanently |

A transient outage (NATS briefly unavailable, node restarting, control-plane-backed minter hiccup) at the wrong moment permanently killed agent startup or dropped workloads.

## Changes

- **`internal/retry` (new):** generic ctx-aware `retry.Do` with exponential backoff + jitter. Lives in its own subpackage because `internal` imports `sdk/go/agent` (import cycle otherwise).
- **Node mints** (`node.go`, `handlers.go`): all 6 sites wrapped. Handler-side retries are short (3 attempts, 100ms→1s) to fit inside the caller's request timeout; startup mints get 5 attempts, 250ms→5s. Resume-state mint failure now logs at Error (workload is dropped) instead of Warn.
- **Agent SDK registration** (`runner.go`): both requests retried (5 attempts, 250ms→5s). Only transport errors retry — an explicit `Success=false` rejection does not.
- **AgentWatcher backoff** (`watcher.go`): 1s doubling to a 30s cap between restart attempts, ctx-aware.
- **Bug fix:** the embedded-agent loop registered the agent as running even when every restart attempt failed — the `restartCount > resetLimit` branch was unreachable inside `for range a.resetLimit`. Failed agents are no longer added to `embeddedAgents`.

## Testing

- `internal/retry` table tests (success paths, exhaustion, ctx cancellation, backoff timing)
- `TestNodeStartSurvivesTransientMintFailures`: node + inmem agent with a minter that fails its first 2 calls of each method — agent still registers, both methods proven retried
- `TestRemoteAgentInitRetriesUntilNodeResponds`: responder appears 400ms after first attempt would have failed with ErrNoResponders — previously failed instantly
- Watcher backoff delay/cancellation tests
- Full `go test ./...` green, including the creds-rotation e2e suite

Stacked on #521 (`hotfix-010-cred-refresh`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -306,7 +306,7 @@ func (n *NexNode) handleAuctionDeployWorkload() func(micro.Request) {
 		// NATS permissions — specifically the log sub scope
 		// $NEX.FEED.<ns>.logs.> — must line up with where the workload
 		// actually lives, which is req.Namespace.
-		wlNatsConn, err := n.minter.Mint(models.WorkloadCred, req.Namespace, workloadID)
+		wlNatsConn, err := n.handlerMinter.Mint(models.WorkloadCred, req.Namespace, workloadID)
 		if err != nil {
 			n.handlerError(r, err, models.ErrCodeInternalServerError, "failed to mint workload nats connection")
 			return
@@ -596,7 +596,7 @@ func (n *NexNode) handleRegisterAgent() func(micro.Request) {
 			return
 		}
 
-		natsConn, err := n.minter.Mint(models.AgentCred, "", agentID)
+		natsConn, err := n.handlerMinter.Mint(models.AgentCred, "", agentID)
 		if err != nil {
 			n.handlerError(r, err, models.ErrCodeInternalServerError, "failed to mint nats connection")
 			return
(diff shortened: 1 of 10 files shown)


Jarema:
Deep multi-agent review of this PR (8 finder angles, adversarial verification) surfaced 10 findings — all addressed in ed88cbe:

**Correctness (would have shipped broken):**
1. **Non-idempotent registration retry** — retrying on *timeout* could resend a registration the node already processed → `already registered` → agent aborts while actually registered. Now only `nats.ErrNoResponders` retries; everything else is `retry.Permanent`.
2. **~15-minute startup hang worst case** — 5×1min register timeouts nested inside 3 watcher restarts, all holding `WaitForAgents`. Collapsed by fix 1.
3. **WaitGroup leak → permanent `Start()` deadlock** (pre-existing) — bulk `Add(N)` up front + `continue` on mint failure never `Done()`'d. Now `Add(1)` per successfully-minted agent.
4. **`StartLocalBinaryAgent` deadlock** (pre-existing) — all-attempts-fail path never released the startup gate.
5. **`/metrics` re-registration panic** (pre-existing) — watcher re-running `Run` re-registered the handler on the default mux → process crash. Now `sync.Once`.
6. **`agentCount` data race** (pre-existing) — mutated from concurrent goroutines under different locks. Now atomic; watcher tests pass under `-race`.

**Simplification:**
7. Three duplicated retry-tuning const triples → named `retry.Long`/`retry.Short` policies owned by the retry package.
8. Six inline `retry.Do` mint closures → one `credentials.WithRetry` decorator at the `CredVendor` seam (startup + handler instances); future call sites get retry for free.
9. Watcher restart backoff had no jitter (lockstep re-stampede after a shared blip) → shares `retry.Jitter`.
10. `restartBackoffDelay` special case → single clamped-shift expression; embedded restart loop flattened.

**Accepted trade-offs (not fixed, deliberate):** minter retries still retry deterministic config errors (~3.75s extra at startup on a permanently misconfigured seed) — OSS minters are local crypto so the retry only pays off for control-plane-backed vendors, but classification isn't possible at this seam. `RemoteAgentInit` keeps `context.Background()` (no ctx in its public signature); worst case is bounded at ~19s.

Full suite green including creds-rotation e2e and `client` (95s).

piotrpio (dismissed):
LGTM

joeriddles (dismissed):
LGTM

joeriddles (approved):

